PR #69, opened 2025-02-26
fix(guard): fix gitlab caching null

forestileao:
## Description
- Fix inconsistent cache: Since credentials is a map, the is_nil would always return false. Setting empty Crendentials
Tested with these steps:
- Cleared instance_config cache
- Recreated gilab integration
- Created a project

Captura de Tela 2025-02-26 às 14 15 47

Captura de Tela 2025-02-26 às 14 16 17

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Enterprise Edition feature
- [ ] Test updates

## Testing

- [ ] Unit tests added/updated
- [ ] Integration tests added/updated
- [x] Manual testing performed
- [ ] Not applicable

## Documentation

- [ ] Documentation update required
- [ ] Documentation updated
- [x] No documentation changes needed

## Checklist

- [x] Code follows project style guidelines
- [x] Self review performed
- [ ] Comments added to hard-to-understand areas
- [ ] Tests prove change is effective
- [ ] Relevant documentation updated
- [ ] Changelog updated if needed
- [ ] CI/CD changes required

### DIFF
--- a/guard/lib/guard/git_provider_credentials.ex
+++ b/guard/lib/guard/git_provider_credentials.ex
@@ -47,7 +47,8 @@ defmodule Guard.GitProviderCredentials do
     cache_key = "#{provider}_credentials"
 
     case Cachex.get(:config_cache, cache_key) do
-      {:ok, credentials} when not is_nil(credentials) ->
+      {:ok, %{client_id: client_id, client_secret: client_secret} = credentials}
+      when not is_nil(client_id) and not is_nil(client_secret) ->
         {:ok, credentials}
 
       _ ->


pedroforesti:
@radwo There are more issues that I noticed. The cache these configurations are never invalidated after a new git integration is set/deleted/updated

forestileao:
@radwo, Ready for review. Added cache invalidation and tested without the pre-set secrets, everything worked well.

guard/lib/guard/grpc_servers/instance_config_server.ex
```diff
@@ -34,9 +34,11 @@ defmodule Guard.GrpcServers.InstanceConfigServer do
 
   def modify_config(request, _stream) do
     Watchman.benchmark("modify_config.duration", fn ->
-      request = request |> request_atom_type()
+      parsed_request = request |> request_atom_type()
+      response = modify_config_(parsed_request.config)
 
-      modify_config_(request.config)
```
radwo, line 39: you should also remove the definition of this method, if it's no longer used.

radwo (changes requested):
Can you elaborate on why you use RabbitMQ messages when the whole flow is within one service?

If you want to run it in the background, just use a Task.

guard/lib/guard/grpc_servers/instance_config_server.ex
forestileao: The whole flow might be in the same project, but the cache is from guard-id and not guard-instance-config. And since Cachex is a memory cache, it will not be able to clean by just calling Cachex.clean on guard

pedroforesti: > you should also remove the definition of this method, if it's no longer used.

I didn't get it, I am still calling modify_config_ in the lines above (line 38).

radwo: > I didn't get it, I am still calling modify_config_ in the lines above (line 38).

My bad 🙏

radwo (approved):

Issue #35, opened 2025-02-26
Encoded characters in error messages


AleksandarCole:
### Describe the bug

I noticed some encoded spaces in error messages. Check screenshot for more details

### Steps to reproduce

1. Go to any project
2. Go to settings/repository
3. Try to change repo to non-existing one
4. See the error message
5. Change repo successfully
6. Check webhook error message

### Additional context

![Image](https://github.com/user-attachments/assets/163b8b17-5fd7-4c80-bcb1-5ce292f60022)

AleksandarCole:
👍 fixed:

![Image](https://github.com/user-attachments/assets/bb5afb51-8381-4c95-8d2e-27c199af9f27)

PR #106, opened 2025-03-06
fix(front): decode URL encoded grpc error messages

forestileao:
## 📝 Description

Related Task: https://github.com/semaphoreio/semaphore/issues/35

- Some Grpc error messages are being URL encoded to [an update in its grpc library ](https://github.com/elixir-grpc/grpc/pull/271)
- Added Uri encode in grpc responses to ensure it will be decoded. If not, it will keep the same.

image

image

## ✅ Checklist

### DIFF
--- a/front/lib/front_web/controllers/notifications_controller.ex
+++ b/front/lib/front_web/controllers/notifications_controller.ex
@@ -91,12 +91,12 @@ defmodule FrontWeb.NotificationsController do
 
           {:error, %{status: :invalid_argument, message: message}} ->
             conn
-            |> put_flash(:alert, message)
+            |> put_flash(:alert, URI.decode(message))
             |> redirect(to: notifications_path(conn, :new))
 
           {:error, %{status: 3, message: message}} ->
             conn
-            |> put_flash(:alert, message)
+            |> put_flash(:alert, URI.decode(message))
             |> redirect(to: notifications_path(conn, :new))
 
           {:error, %{message: message, status: 9}} ->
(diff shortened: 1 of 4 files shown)


radwo (commented):
I think it would be better to fix this issue at the source; in the repositoryhub. That way, it will automatically work everywhere it’s used, including in places like the Insider.

forestileao:
> I think it would be better to fix this issue at the source; in the repositoryhub. That way, it will automatically work everywhere it’s used, including in places like the Insider.
This is becoming standard for new grpc servers. As mentioned in the description of the PR, this encoding was implemented as a fix directly in the library to follow the standard for Grpc responses.

radwo:
Oh, now I see 👍 
Did you consider updating the `grpc` to `0.6.0` in front so that the lib will handle decoding?

forestileao:
> Oh, now I see 👍 
> Did you consider updating the `grpc` to `0.6.0` in front so that the lib will handle decoding?

Good point. Let me  try updating, probably it will solve everything.

pedroforesti:
> Oh, now I see 👍 Did you consider updating the `grpc` to `0.6.0` in front so that the lib will handle decoding?

@radwo Looks like to make this update we will need to also update protobuff. And  we would need to update every single proto struct in front, or we are going to face warnings as below (what conflicts with our warning as error police for releases):

```sh
warning: InternalApi.Projecthub.RequestMeta.new/1 is deprecated. Build the struct by hand with %MyMessage{...} or use struct/2
17.90 Invalid call found at 10 locations
```

I will create another Issue for this and reference the this thread.

radwo (approved):

hamir-suspect (approved):

forestileao:
This will be merged as a workaround, but the issue for upgrading the grpc version is created [here](https://github.com/semaphoreio/semaphore/issues/114), and it is already assigned to me.

Issue #30, opened 2025-02-26
Emphasise job status on job page


AleksandarCole:
### Is your feature request related to a problem? Please describe.

Currently if the job fails the only indicator on the job page is the failed command. 

In some cases, that may not be enough to communicate the job state. Check the screenshot bellow. Since job has test reports configured and tests are passing, user can easily be under impression that everything is ok.

![Image](https://github.com/user-attachments/assets/423286d4-c659-46ab-907a-05f350475d41)

### Describe the solution you'd like

A new fail/pass indicator should be added next to the job name in the header of the page, to communicate the state of the job more clearly.

PR #120, opened 2025-03-11
feat(front): add job status badge

forestileao:
## 📝 Description
- Related Issue: https://github.com/semaphoreio/semaphore/issues/30
- Add job status badge next to its title to show its status in real time

image

image

## ✅ Checklist

### DIFF
--- a/guard/lib/guard/front_repo.ex
+++ b/guard/lib/guard/front_repo.ex
@@ -3,18 +3,6 @@ defmodule Guard.FrontRepo do
     otp_app: :guard,
     adapter: Ecto.Adapters.Postgres
 
-  defmodule Quota do
-    use Ecto.Schema
-
-    @primary_key {:id, :binary_id, autogenerate: true}
-
-    schema "quotas" do
-      field(:type, :string)
-      field(:value, :integer)
-      field(:organization_id, :binary_id)
-    end
-  end
-
   defmodule OauthConnection do
     use Ecto.Schema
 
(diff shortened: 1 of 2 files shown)


front/lib/front_web/templates/job/_status_badge.html.eex
```diff
@@ -0,0 +1,8 @@
+<div
+  class="flex mt1"
+  data-poll-background
+  data-poll-state="<%= @badge_pollman.state %>"
+  data-poll-href="<%= @badge_pollman.href %>"
+>
+  <%= raw job_status_badge(@job) %>
```
hamir-suspect: instead of `raw job_status_badge` please put here the `<span>` from the view and call view helper just for for the class.

front/lib/front_web/views/job_view.ex
```diff
@@ -62,4 +62,16 @@ defmodule FrontWeb.JobView do
...
     end
   end
+
+  def job_status_badge(nil), do: ""
+
+  def job_status_badge(job),
+    do:
+      "<span class='#{job_state_color(job.state)} white br1 ph2'>#{String.capitalize(job.state)}</span>"
+
+  defp job_state_color("pending"), do: "bg-orange"
```
hamir-suspect, line 66: this is fine to stay here

front/lib/front_web/views/job_view.ex
```diff
@@ -62,4 +62,16 @@ defmodule FrontWeb.JobView do
       "The job is waiting on #{job.machine_type} quota."
     end
   end
+
+  def job_status_badge(nil), do: ""
+
+  def job_status_badge(job),
```
hamir-suspect: I don't think this should be here

hamir-suspect (changes requested):

hamir-suspect (approved):

Issue #86, opened 2025-03-04
Restore Refresh button on people page


AleksandarCole:
### Is your improvement proposal related to a problem? Please describe.

On Semaphore Cloud on the page for adding people to organization there was a `Refresh` button that forced the check to re-sync access/permissions within Semaphore organization with access on GitHub repositories. 

When redesign of the UI for adding people to organization was pushed, this button, and functionality was lost. 

### Describe the solution you'd like

Add a new button to the organization people page. Design submitted bellow. 
Button should have the same behaviour as a previous `Refresh` button, no new functionality is needed. 
___
<img width="1488" alt="Image" src="https://github.com/user-attachments/assets/c473590c-2a32-45d7-bd39-4da911485543" />

___

UI design: [app-design: c70c517](https://github.com/renderedtext/app-design/commit/c70c517e2b8c08f51c8e8db06a755b962b533bae#diff-1ce8b379dfa62b4ad619de291392c11e9202ea2854b333530a16c5a84ebb6d8cR38-R40)

pedroforesti:
The feature is merged and I can verify that in CE the feature is disabled, but it appears on prod as planned:

CE:
![Image](https://github.com/user-attachments/assets/8a871a61-b0b2-4147-aabc-7c532c73fd1d)

Cloud

![Image](https://github.com/user-attachments/assets/d7dd4c16-8ca2-4577-b7c2-ceca033e3ba5)

PR #124, opened 2025-03-13
Fix/front/hide refresh button in ce

forestileao:
## 📝 Description
Related Issue: https://github.com/semaphoreio/semaphore/issues/86

- The sync people button should only be enabled in EE
- Rbac refreshCollaborators only exists in RBAC EE

image
 

## ✅ Checklist

### DIFF
--- a/front/config/runtime.exs
+++ b/front/config/runtime.exs
@@ -171,6 +171,10 @@ config :front,
        System.get_env("HIDE_PROMOTIONS") == "true"
 
 config :front,
+       :hide_refresh_collaborators,
+       System.get_env("HIDE_REFRESH_COLLABORATORS") == "true"
+
+config :front,
        :workflow_templates_path,
        System.fetch_env!("WORKFLOW_TEMPLATES_YAMLS_PATH")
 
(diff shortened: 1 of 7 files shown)


front/config/runtime.exs
```diff
@@ -170,6 +170,10 @@ config :front,
        :hide_promotions,
        System.get_env("HIDE_PROMOTIONS") == "true"
 
+config :front,
+       :hide_refresh_collaborators,
```
skipi: @pedroforesti I'm wondering if this is a way to go 🤔 

WDYT about using a `ui_show_refresh_collaborators` feature flag instead? By default, unknown feature flags are `disabled`.

skipi (changes requested):

front/config/runtime.exs
pedroforesti: I think it is good to have this control to be able to disable or enable it when we want, so let me implement it.  I had implemented this env variable based on hide_promotions.
But that is no problem. Let's follow the feature flag approach.

skipi (approved):

Issue #27, opened 2025-02-26
Change YAML Editor


AleksandarCole:
### Is your feature request related to a problem? Please describe.

Currently, we have two different YAML editors in the project creation screen (when user is selecting the starter template), and the Edit workflow screen (when user is editing the yaml). 

The one that we use for Edit workflow is outdated and also, it strips yaml comments.

### Describe the solution you'd like

1. Make sure we are using single yaml editor (update the one on Edit workflow screen)
2. ~Ensure that we are not stripping yaml comments when editing yaml in app~ - REMOVED FROM SCOPE

lucianoliberti:
Waiting for response/next steps [here](https://discord.com/channels/1097422014735732746/1344321610861252659/1349436389678387200).

cc @AleksandarCole @forestileao

forestileao:
@AleksandarCole I think the code editor can be changed to the new one, but the comments will still be stripped because of the problem described in https://github.com/semaphoreio/semaphore/issues/28#issuecomment-2719464016

forestileao:
The feature is merged and I can verify that it is working with validation as well:

https://github.com/user-attachments/assets/06341807-77d9-49c6-a54e-37b8338a68c3

PR #126, opened 2025-03-14
refactor(front): add new yaml editor to workflow builder

forestileao:
## 📝 Description
Issue: https://github.com/semaphoreio/semaphore/issues/27

- Add new monaco editor in code workflow editor
- The feature is behind ui_new_workflow_code_editor feature flag
- Add yaml Schema validation for the pipelines
Captura de Tela 2025-03-14 às 15 02 14

https://github.com/user-attachments/assets/a1f7b04e-d722-4258-b684-92c4fb8dd53f

## ✅ Checklist


front/assets/js/workflow_editor/components/new_code_editor.js
```diff
@@ -0,0 +1,114 @@
+import { render } from 'preact';
```
radwo: How about renaming this to `monaco_code_editor.js`?
`new_` doesn’t add any meaning, and if another rewrite happens, it could be confusing since it will no longer be `new`.

forestileao:
@skipi Since the schema validation required the yaml editor. I merged this feature in this same PR.
https://github.com/user-attachments/assets/a59a924c-dac8-46e2-8bd9-f164f0591a3c

skipi (dismissed):

DamjanBecirovic:
@pedroforesti how do we get this compiled version of schema in the `front/assets/js/workflow_editor/models/pipeline_schema_validator.js`  file?

Can you add instructions or, even better, a Makefile target in the `plumber/spec` directory to generate the new version of that file when we change the yaml schema?

forestileao:
> @pedroforesti how do we get this compiled version of schema in the `front/assets/js/workflow_editor/models/pipeline_schema_validator.js`  file?
> 
> Can you add instructions or, even better, a Makefile target in the `plumber/spec` directory to generate the new version of that file when we change the yaml schema?

@DamjanBecirovic that's a good ideia! It's a cli command from Ajv that generates it. I will probably need to generate a script for some specific parsing, but that's okay. Working on it.

helm-chart/templates/configmaps/features.yaml
```diff
@@ -98,6 +98,10 @@ data:
       enabled: true
     ui_agent_page:
       enabled: true
+    ui_show_refresh_collaborators:
```
lucaspin, line 101: Why was `ui_show_refresh_collaborators ` added here? It doesn't seem related to this change

pedroforesti: It was missing and should be added at least for EE. I noticed now that it was required to add ui_monaco_workflow_code_editor to this file. So I just wanted to get the opportunity and add ui_show_refresh_collaborators as well.

lucaspin (approved):

lucaspin (approved):

VeljkoMaksimovic (approved):

PR #130, opened 2025-03-14
feat(velocity): enable velocity in ee

dexyk:
## 📝 Description

Enables velocity in EE.

For more details, see: [project task](https://github.com/renderedtext/project-tasks/issues/2276)

## ✅ Checklist
- [x] ~This change requires documentation update~ - N/A

### DIFF
--- a/front/helm/templates/job-page.yaml
+++ b/front/helm/templates/job-page.yaml
@@ -120,8 +120,10 @@ spec:
               value: {{ .Values.gettingStarted.mountPath | quote }}
             - name: WORKFLOW_TEMPLATES_YAMLS_PATH
               value: {{ .Values.workflowTemplates.mountPath | quote }}
+{{- if eq .Values.global.edition "ce" }}
             - name: "SKIP_VELOCITY"
               value: "true"
+{{- end }}
             - name: "HIDE_PROMOTIONS"
               value: "true"
             - name: "CE_ROLES"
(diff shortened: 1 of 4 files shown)


skipi (approved):

front/helm/templates/job-page.yaml
```diff
@@ -120,8 +120,10 @@ spec:
               value: {{ .Values.gettingStarted.mountPath | quote }}
             - name: WORKFLOW_TEMPLATES_YAMLS_PATH
               value: {{ .Values.workflowTemplates.mountPath | quote }}
+{{- if eq .Values.global.edition "ce" }}
             - name: "SKIP_VELOCITY"
```
lucaspin, line 124: One thing here: we have a lot of environment variables (SKIP_VELOCITY, HIDE_PROMOTIONS, CE_ROLES, ...) that are included or not based on `.Values.global.edition`. It's probably simpler to just pass a single `EDITION` environment variable instead of all of those

dexyk: @lucaspin you mean replacing all these environment variables that internally control the front behavior with a single value, or just importing them from a single location? If we replace them with a single env. var., we lose configurability of the service.

lucaspin: I mean replacing them with a single var. Yes, we lose configurability of the service, but configurability makes things more complicated. Those variables are only gonna be tweaked based on the edition, no?

dexyk: @lucaspin I would rather keep them like this at the moment, and reduce later down the line if we decide to do so

lucaspin (approved):

PR #142, opened 2025-03-19
feat(rbac): jit saml group membership

VeljkoMaksimovic:
## 📝 Description

Handle group membership for jit saml users based on saml attributes

https://github.com/renderedtext/tasks/issues/7653

## ✅ Checklist


ee/rbac/lib/rbac/okta/saml/jit_provisioner/add_user.ex
```diff
@@ -125,6 +127,29 @@ defmodule Rbac.Okta.Saml.JitProvisioner.AddUser do
...
+      e ->
+        info("[Saml JIT Provisioner] Response from IdpGroupMapping.map_groups #{inspect(e)}")
+        {:ok, []}
+    end
+  end
+
+  defp assign_groups(user_id, group_ids) when is_list(group_ids) do
+    Enum.each(group_ids, fn group_id ->
+      Rbac.Repo.GroupManagementRequest.create_new_request(user_id, group_id, :add)
+    end)
```
pedroforesti, line 148: Is it expected to receive a large number of groups? Just thinking if there is this scenario, then we could use a Task async stream or something. But if it is limited to just a few groups, then it would not be required.

pedroforesti (approved):
Reviewed but with a suggestion of Improvement IF APPLICABLE

forestileao (approved):

PR #146, opened 2025-03-20
fix(rbaac): SAML assigned roles can be managed manually

VeljkoMaksimovic:
## 📝 Description

Roles that have been assigned through saml jit provisioning should be managed later on manually. Hence we have to mark them as :manually_assigned

https://github.com/renderedtext/tasks/issues/7653

## ✅ Checklist

### DIFF
--- a/ee/rbac/lib/rbac/okta/saml/jit_provisioner/add_user.ex
+++ b/ee/rbac/lib/rbac/okta/saml/jit_provisioner/add_user.ex
@@ -116,7 +116,9 @@ defmodule Rbac.Okta.Saml.JitProvisioner.AddUser do
       :ok
     else
       {:ok, rbi} = RoleBindingIdentification.new(user_id: user_id, org_id: org_id)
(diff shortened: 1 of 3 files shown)


pedroforesti (approved):

ee/rbac/lib/rbac/okta/saml/jit_provisioner/add_user.ex
```diff
@@ -116,7 +116,9 @@ defmodule Rbac.Okta.Saml.JitProvisioner.AddUser do
       :ok
     else
       {:ok, rbi} = RoleBindingIdentification.new(user_id: user_id, org_id: org_id)
-      {:ok, nil} = RoleManagement.assign_role(rbi, role_id, :saml_jit)
+      # Although this role is assigned through :saml_jit, it can be updated and
```
forestileao:
```suggestion
      # Although this role is assigned through saml_jit, it can be updated and
```

Maybe we could remove this as atom reference since we are not going to have it anymore

forestileao (approved):

forestileao (approved):

PR #152, opened 2025-03-20
fix(github_hooks): raise unknown error on unhandled octokit exception

lucaspin:
## 📝 Description

When we receive an error from Octokit that we are not currently handling, the `pull_request` method on the repo host client returns nil, which can lead to a `undefined method `[]' for nil` error on the hook handler. See [this comment](https://github.com/renderedtext/tasks/issues/7601#issuecomment-2729445192) for more details.

Instead, to help us troubleshoot issues, we should log the actual error message that we are receiving from GitHub.

## ✅ Checklist


radwo (commented):
here is the yet another place where `update_pr_data` is used https://github.com/semaphoreio/semaphore/blob/main/github_hooks/lib/internal_api/repo_proxy/pr_payload.rb#L11
this one is used in the API call, when you would like to start a workflow from a PR.

lucaspin:
@radwo added a rescue there as well

VeljkoMaksimovic (approved):

Issue #135, opened 2025-03-18
GitLab connection keeps disconnecting for more than one pod replica


forestileao:
### Describe the bug

Since Git provider's token caches are not the same for every pod in Guard (it is a memory cache). This can make us get an invalid token when the request reaches a different pod when we have more than 1 replica running for Guard. Also,  an invalid token might be cached for every other pod after the invalidation.

The solution would be: Instead of caching the token in memory for every pod, with the expiry date,  change it to store the expiry date within the RepoHostAccount table. Solving its inconsistency

### Steps to reproduce

1. Go to Profiles Setting
2. If your organization has GitLab feature enabled, refresh the page
3. The GitLab Connection will disconnect and connect every time you refresh the page
4. If you try to create a project, it will fail

### Additional context

This issue just happens if you have more than one Guard Replica running

pedroforesti:
The fix is working and deployed to production. I can verify that the GitLab or BitBucket is not disconnecting by refreshing the profile settings page and calling refresh_repository_token:

<img width="530" alt="Image" src="https://github.com/user-attachments/assets/2a0bd3bd-46a3-457b-9cf1-ee81f9aa0dc7" />

PR #153, opened 2025-03-21
fix(guard): remove token cache from repo host account

forestileao:
## 📝 Description
Related Issue: https://github.com/semaphoreio/semaphore/issues/135

Since Git provider's token caches are not the same for every pod in Guard (it is a memory cache). This can make us get an invalid token when the request reaches a different pod when we have more than 1 replica running for Guard. Also, an invalid token might be cached for every other pod after the invalidation.

To fix this, let's add a new column called token_expires_at and make this validation directly within the database data. Since we already fetch it anyway.

## ✅ Checklist


guard/lib/guard/utils.ex
```diff
@@ -58,20 +58,24 @@ defmodule Guard.Utils.OAuth do
     refresh_token = body["refresh_token"]
 
     expires_at = calc_expires_at(expires_in)
+    nil_valid = repo_host_account.repo_host == "github"
```
lucaspin, line 63: Why do we allow it to be empty for GH, but not for the other providers?

pedroforesti: This needs to be documented, so I'll add it to the code ;).

By default, GitHub tokens do not expire unless the optional setting in the GitHub app is changed. If altered, the `expires_in` and `refresh_token` will not be null. That's why we have refresh token logic for GitHub, even if it's not typically used.

lucaspin (approved):

VeljkoMaksimovic (approved):

PR #155, opened 2025-03-21
feat(front): start a job to commit changes from workflow editor

DamjanBecirovic:
## 📝 Description

Introduces an option to commit changes made in the workflow editor through a Semaphore job that will use git CLI for these actions. The previous approach relies on libgit2 and API actions which seem to be unstable for large git repositories.

## ✅ Checklist
- ~[x] This change requires documentation update~ - N/A


skipi (approved):

zebra/lib/zebra/apis/public_job_api/serializer.ex
```diff
@@ -27,6 +27,17 @@ defmodule Zebra.Apis.PublicJobApi.Serializer do
   end
 
   def spec(job) do
+    # for restricted jobs we do not want to expose their environment
+    if job.spec["restricted_job"] do
```
lucaspin, line 31: Would it make more sense to hide those jobs completely in the API?

zebra/test/zebra/apis/public_job_api_test.exs
```diff
@@ -282,6 +337,56 @@ defmodule Zebra.Api.PublicJobApiTest do
 
       assert reply == %GRPC.RPCError{message: "Job #{job.id} not found", status: 5}
     end
+
+    test "if a job is restircted do not return configuration in spec" do
```
lucaspin:
```suggestion
    test "if a job is restricted do not return configuration in spec" do
```

zebra/test/zebra/apis/public_job_api_test.exs
```diff
@@ -185,6 +185,61 @@ defmodule Zebra.Api.PublicJobApiTest do
 
       assert Enum.map(reply.jobs, & &1.metadata.id) == [job1.id]
     end
+
+    test "if a job is restircted do not return configuration in spec" do
```
lucaspin:
```suggestion
    test "if a job is restricted do not return configuration in spec" do
```

zebra/lib/zebra/apis/public_job_api/serializer.ex
DamjanBecirovic: They are occupying resources and quotas and are not visible in UI except in the Activity monitor while running, so I'm not sure if it is ok to hide them completely. 🤔

lucaspin: Fair enough

pedroforesti (approved):

forestileao (approved):

PR #168, opened 2025-03-25
fix(front: refactor timestamps format to up to 1 hour

forestileao:
## 📝 Description
As discussed with @AleksandarCole, we introduced a new rule to the timestamps:
- If the time difference is less than 1 hour, it will be displayed as a relative time (e.g., "5 minutes ago").
- If the time difference is between 1 hour and 3 days, it will now follow the format: {weekday} #{suffixed_day} {month} {year} at HH:mm (e.g., "Tue 5th Mar 2024 at 14:30").
- If the time difference is greater than 3 days, it will follow the existing full date format (e.g., "on Tue 5th Mar 2024").
image

## ✅ Checklist

### DIFF
--- a/front/test/front/utils_test.exs
+++ b/front/test/front/utils_test.exs
@@ -79,10 +79,10 @@ defmodule Front.UtilsTest do
     assert Front.Utils.decorate_relative(thirty_minutes_ago) == "30 minutes ago"
 
     assert Front.Utils.decorate_relative(fake_today, two_days_ago) ==
-             "on Wed 18th Mar 2025 at 14:30"
+             "on Tue 18th Mar 2025 at 14:30"
 
     assert Front.Utils.decorate_relative(fake_today, three_days_ago) ==
-             "on Tue 19th Mar 2025 at 10:15"
+             "on Mon 17th Mar 2025 at 10:15"
 
     assert Front.Utils.decorate_relative(~U[2025-03-05 22:05:26.833945Z]) ==
              "on Wed 05th Mar 2025"
--- a/front/test/front/utils_test.exs
+++ b/front/test/front/utils_test.exs
@@ -64,15 +64,33 @@ defmodule Front.UtilsTest do
   end
 
   test "decorate relative" do
-    two_days_ago = Timex.now() |> Timex.shift(days: -2) |> Timex.to_unix()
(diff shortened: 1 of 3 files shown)


front/test/front/utils_test.exs
```diff
@@ -64,15 +64,33 @@ defmodule Front.UtilsTest do
...
-    assert Front.Utils.decorate_relative(DateTime.utc_now()) == "now"
-    assert Front.Utils.decorate_relative(two_days_ago) == "2 days ago"
-    assert Front.Utils.decorate_relative(three_days_ago) == "3 days ago"
+    assert Front.Utils.decorate_relative(now) == "now"
+
+    assert Front.Utils.decorate_relative(thirty_minutes_ago) == "30 minutes ago"
+
+    ordinal_suffix_regex = "(st|nd|rd|th)"
+
+    assert Regex.match?(
```
skipi: @forestileao - what do you think about improving the readability of these tests by asserting actual string values (e.g., `"on Wed 05th Mar 2025"`)?

pedroforesti: In [here](https://github.com/semaphoreio/semaphore/pull/168/files/29a5163f071bc3098e03f8271f4e3c358d20b157#diff-ef5721660b4f38b0c4c8cdf683dbbd16a59835b44c04716654897b8f8c18c2b2) it makes sense. So I will change.

But for the others is a different case, because I want to test if "Two Days ago" follows the correct format. 

If the time difference is between 1 hour and 3 days, it will now follow the format: {weekday} #{suffixed_day} {month} {year} at HH:mm (e.g., "Tue 5th Mar 2024 at 14:30").

So if I put two days ago and put the actual string, after some days this test will fail, since it will not follow the relative format.

skipi: Ah - I see - the `decorate_relative/1` uses `now`. How hard would it be to pass two dates there? It'll make testing much easier. We can always preserve the old behavior with:

```elixir
def decorate_relative(date), do: decorate_relative(DateTime.utc_now(), date)
def decorate_relative(from, until) do 
  #stuff
end
```

pedroforesti: that is an excelent idea, Let me implement it :)

skipi (approved):

PR #205, opened 2025-04-07
refactor(guard): do not remove org and projects deletion after user deletion

forestileao:
## 📝 Description
For security reasons and because a new soft-deletion will be implemented soon, we are deleting the code that removes the organizations and projects after a user is deleted.

## ✅ Checklist
- [ ] This change requires a documentation update

### DIFF
--- a/guard/lib/guard/store/user.ex
+++ b/guard/lib/guard/store/user.ex
@@ -246,41 +246,14 @@ defmodule Guard.Store.User do
 
     def delete_with_owned_orgs(user_id) do
       result =
-        Ecto.Multi.new()
-        |> Ecto.Multi.run(:get_user, fn repo, _changes ->
-          case repo.get(Repo.User, user_id) do
-            nil -> {:error, :user_not_found}
-            user -> {:ok, user}
-          end
-        end)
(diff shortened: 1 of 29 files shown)


VeljkoMaksimovic (approved):

lucaspin (approved):

guard/lib/guard/store/user.ex
```diff
@@ -254,27 +254,19 @@ defmodule Guard.Store.User do
...
-            :ok = Guard.Api.Project.destroy_all_projects_by_org_id(org.id)
+          repo.delete_all(from(r in Repo.RepoHostAccount, where: r.user_id == ^user_id))
 
-            repo.delete_all(from(m in Repo.Member, where: m.organization_id == ^org.id))
-
-            repo.delete_all(
-              from(s in Repo.OrganizationSuspension, where: s.organization_id == ^org.id)
+          members_query =
+            from(m in Guard.FrontRepo.Member,
+              join: rha in Guard.FrontRepo.RepoHostAccount,
```
VeljkoMaksimovic, line 259: Will it have anything to join with if we delete all the repo_host_accounts on the line 257?

VeljkoMaksimovic (approved):

PR #209, opened 2025-04-09
feat(projecthub): create soft delete for projects

forestileao:
## 📝 Description
As a plan for making projects restorable. We are implementing a soft-delete for projects.
In another PR, a garbage collector for projects will be implemented in project_hub.

## Main changes:
- Project Describe and DescribeMany only searches for projects where deleted_at is null
- Created a Deleted By and Deleted At column to store who is the requester of the deletion and when it was deleted for the GarbageCollector
-  Created a new routing key called soft_deleted for project exchange. So the cache will be invalidated in front.

## ✅ Checklist

### DIFF
--- a/front/lib/front/layout/cache_invalidator.ex
+++ b/front/lib/front/layout/cache_invalidator.ex
@@ -36,7 +36,6 @@ defmodule Front.Layout.CacheInvalidator do
       {"organization_exchange", "suspension_removed", :suspension_removed},
       {"project_exchange", "created", :created_project},
       {"project_exchange", "updated", :updated_project},
-      {"project_exchange", "deleted", :deleted_project},
       {"project_exchange", "soft_deleted", :deleted_project},
       {"user_exchange", "favorite_created", :starred},
       {"user_exchange", "favorite_deleted", :unstarred}
--- a/front/lib/front/layout/cache_invalidator.ex
+++ b/front/lib/front/layout/cache_invalidator.ex
@@ -37,6 +37,7 @@ defmodule Front.Layout.CacheInvalidator do
       {"project_exchange", "created", :created_project},
       {"project_exchange", "updated", :updated_project},
       {"project_exchange", "deleted", :deleted_project},
+      {"project_exchange", "soft_deleted", :deleted_project},
       {"user_exchange", "favorite_created", :starred},
       {"user_exchange", "favorite_deleted", :unstarred}
     ]
--- a/projecthub/lib/projecthub/events/project_deleted.ex
+++ b/projecthub/lib/projecthub/events/project_deleted.ex
@@ -1,5 +1,8 @@
 defmodule Projecthub.Events.ProjectDeleted do
-  def publish(project) do
+  @soft_deleted_routing_key "soft_deleted"
+  @deleted_routing_key "deleted"
+
+  def publish(project, opts \\ []) do
     timestamp = DateTime.utc_now() |> DateTime.to_unix()
 
     event =
@@ -10,11 +13,12 @@ defmodule Projecthub.Events.ProjectDeleted do
       )
 
     message = InternalApi.Projecthub.ProjectDeleted.encode(event)
+    routing_key = if opts[:soft_delete], do: @soft_deleted_routing_key, else: @deleted_routing_key
 
     options = %{
       url: Application.fetch_env!(:projecthub, :amqp_url),
       exchange: "project_exchange",
-      routing_key: "deleted"
+      routing_key: routing_key
     }
 
     Tackle.publish(message, options)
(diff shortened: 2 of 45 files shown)


projecthub/lib/projecthub/models/project.ex
```diff
@@ -251,7 +256,7 @@ defmodule Projecthub.Models.Project do
 
   def find(id) do
     if id_is_uuid?(id) do
-      case Repo.get_by(Project, id: id) do
+      case from(Project) |> where([p], p.id == ^id and is_nil(p.deleted_at)) |> Repo.one() do
```
VeljkoMaksimovic:
```suggestion
      case from(Project) |> where([p], p.id == ^id) |> where_undeleted() |> Repo.one() do
```

front/lib/front/layout/cache_invalidator.ex
```diff
@@ -37,6 +37,7 @@ defmodule Front.Layout.CacheInvalidator do
       {"project_exchange", "created", :created_project},
       {"project_exchange", "updated", :updated_project},
       {"project_exchange", "deleted", :deleted_project},
```
VeljkoMaksimovic:
```suggestion
```

If I understand correctly, this event will not be published anymore, correct?

projecthub/lib/projecthub/workers/project_init.ex
```diff
@@ -61,6 +61,7 @@ defmodule Projecthub.Workers.ProjectInit do
       Watchman.benchmark("#{@metric_name}.tick.duration", fn ->
         Projecthub.Models.Project
         |> where([p], p.state == ^StateMachine.initializing() or p.state == ^StateMachine.initializing_skip())
+        |> where([p], is_nil(p.deleted_at))
```
VeljkoMaksimovic: Why do we need this? Will it be possible for projects that are yet to be created to already have deleted_at?

projecthub/lib/projecthub/workers/project_init.ex
```diff
@@ -93,6 +94,7 @@ defmodule Projecthub.Workers.ProjectInit do
           Projecthub.Models.Project
           |> where([p], p.state == ^StateMachine.initializing() or p.state == ^StateMachine.initializing_skip())
           |> where([p], p.id == ^project_id)
+          |> where([p], is_nil(p.deleted_at))
```
VeljkoMaksimovic: Same comment as above

front/lib/front/layout/cache_invalidator.ex
VeljkoMaksimovic: Correction, now I see that the deleted event will be used to communicate from projecthub to artifacthub, secretshub, etc. but I think the question is still valid: Is there anything to refresh in the UI once the project is removed by the garbage collectior/worker, or are we only making UI changes once the project is soft deleted?

VeljkoMaksimovic (changes requested):

front/lib/front/layout/cache_invalidator.ex
forestileao: You are right, we can only listen by soft_deleted

projecthub/lib/projecthub/workers/project_init.ex
forestileao: You are correct, there is no need for this in project_init

VeljkoMaksimovic (approved):

projecthub/lib/projecthub/events/project_deleted.ex
```diff
@@ -10,11 +13,12 @@ defmodule Projecthub.Events.ProjectDeleted do
       )
 
     message = InternalApi.Projecthub.ProjectDeleted.encode(event)
+    routing_key = if opts[:soft_delete], do: @soft_deleted_routing_key, else: @deleted_routing_key
```
hamir-suspect, line 16: In what cases are we keeping the `deleted_routing_key` for this message? 
If we will still have cases with the routing key of this message being `"deleted"` the [front layout cache invalidator ](https://github.com/semaphoreio/semaphore/pull/209/files#diff-73faa1847a2958311e5e56f7e923acc7a4abdaea6d2d4bbd6ea2e8f2a8d8beeaL39) still has to react to messages on `"deleted"` routing key as well as to the `"soft_deleted"`

hamir-suspect (changes requested):

projecthub/lib/projecthub/events/project_deleted.ex
pedroforesti: It will be necessary to have it once I implement the Garbage Collector of projects and everything that we remove should trigger the original consumers that consume from "deleted". 
Since we just want to trigger them for hard_deletions.

hamir-suspect (approved):

PR #211, opened 2025-04-09
Api/rbac remove user from org

VeljkoMaksimovic:
## 📝 Description

Http api for removing a user from the org (clevertap): https://github.com/renderedtext/tasks/issues/7779

Changes made to RBAC are only here so that the endpoint for retracting roles would not crash if the user is already not part of the organization

I made changes to the plumber Dockerfile, Makefile and test_helper in order to be able to run tests both locally (using make test.ex) and on CI

(tested pipelines API as well, it works)

## ✅ Checklist
- [x] This change requires documentation update (for now we will only have private docs for clevertap)

### DIFF
--- a/public-api/v1alpha/lib/pipelines_api/role_bindings/delete.ex
+++ b/public-api/v1alpha/lib/pipelines_api/role_bindings/delete.ex
@@ -1,6 +1,6 @@
 defmodule PipelinesAPI.RoleBindings.Delete do
   @moduledoc """
-  Plug deletes a deployment target.
+  Plug deletes a role binding (removes role from the subject).
   """
   use Plug.Builder
   require Logger
@@ -107,7 +107,7 @@ defmodule PipelinesAPI.RoleBindings.Delete do
 
   defp lookup_user_by_email(conn, email) do
     case UserApiClient.describe_by_email(email) do
-      {:ok, user} when not is_nil(user) ->
+      {:ok, user} ->
         conn |> Plug.Conn.assign(:user_id, user.id)
 
       {:error, :not_found} ->
--- a/public-api/v1alpha/lib/pipelines_api/role_bindings/delete.ex
+++ b/public-api/v1alpha/lib/pipelines_api/role_bindings/delete.ex
@@ -79,36 +79,44 @@ defmodule PipelinesAPI.RoleBindings.Delete do
 
     cond do
       not is_nil(user_id) ->
-        case UserApiClient.describe_many([user_id]) do
-          {:ok, %{users: users}} when is_list(users) and length(users) > 0 ->
-            [user | _] = users
-            conn |> Plug.Conn.assign(:user_id, user.id)
-
-          {:ok, _} ->
-            Logger.info("User with id #{user_id} not found")
-            Common.respond({:error, {:user, "User with id #{user_id} not found"}}, conn)
-
-          error ->
-            Logger.error("Error when fetching user by ID: #{inspect(error)}")
-            Common.respond({:error, {:internal, "Internal error"}}, conn)
-        end
+        lookup_user_by_id(conn, user_id)
 
       not is_nil(email) ->
-        case UserApiClient.describe_by_email(email) do
-          {:ok, user} when not is_nil(user) ->
-            conn |> Plug.Conn.assign(:user_id, user.id)
-
-          {:error, :not_found} ->
-            Logger.info("User with email #{email} not found")
-            Common.respond({:error, {:user, "User with email #{email} not found"}}, conn)
-
-          error ->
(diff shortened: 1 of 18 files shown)


public-api/v1alpha/lib/pipelines_api/role_bindings/delete.ex
```diff
@@ -0,0 +1,122 @@
...
+      not is_nil(email) ->
+        lookup_user_by_email(conn, email)
+
+      true ->
+        Common.respond({:error, {:user, "Missing user_id or email in query parameters"}}, conn)
+    end
+  end
+
+  defp lookup_user_by_id(conn, user_id) do
+    case UserApiClient.describe_many([user_id]) do
```
lucaspin, line 95: Why are we using describe_many instead of describe?

public-api/v1alpha/lib/pipelines_api/router.ex
```diff
@@ -98,6 +99,8 @@ defmodule PipelinesAPI.Router do
 
   match("/deployment_targets/:target_id", via: :delete, to: DeleteDeploymentTarget)
 
+  match("/rbac/", via: :delete, to: DeleteRoleBindings)
```
lucaspin: Using /rbac/ here for the path could be confusing. What if we need to add a way to delete a custom role or a group here too?

public-api/v1alpha/lib/pipelines_api/role_bindings/delete.ex
```diff
@@ -0,0 +1,122 @@
+defmodule PipelinesAPI.RoleBindings.Delete do
+  @moduledoc """
+  Plug deletes a deployment target.
```
dexyk: @VeljkoMaksimovic please update the doc

public-api/v1alpha/lib/pipelines_api/role_bindings/delete.ex
```diff
@@ -0,0 +1,122 @@
...
+
+      error ->
+        Logger.error("Error when fetching user by ID: #{inspect(error)}")
+        Common.respond({:error, {:internal, "Internal error"}}, conn)
+    end
+  end
+
+  defp lookup_user_by_email(conn, email) do
+    case UserApiClient.describe_by_email(email) do
+      {:ok, user} when not is_nil(user) ->
```
dexyk: @VeljkoMaksimovic is this case possible, that retrieved user is `nil`? If it's possible then we haven't covered that case

public-api/v1alpha/lib/pipelines_api/router.ex
dexyk: @VeljkoMaksimovic I agree with @lucaspin

public-api/v1alpha/lib/pipelines_api/role_bindings/delete.ex
VeljkoMaksimovic: There was existant userApiClient for describe_many, so instead of creating a new one for describe I decided to use this one

public-api/v1alpha/lib/pipelines_api/router.ex
VeljkoMaksimovic: Good catch, changed it to /role_bindings

public-api/v1alpha/lib/pipelines_api/role_bindings/delete.ex
VeljkoMaksimovic: Thanks, updated

public-api/v1alpha/lib/pipelines_api/role_bindings/delete.ex
VeljkoMaksimovic: Correct, it is not possible. Removed the guard

public-api/v1alpha/lib/pipelines_api/role_bindings/delete.ex
```diff
@@ -0,0 +1,122 @@
...
+  plug(:authorize)
+  plug(:fetch_user_to_be_deleted)
+  plug(:delete)
+
+  def delete(conn, _opts) do
+    Metrics.benchmark("PipelinesAPI.router", ["rbac_delete"], fn ->
+      requester_id = conn.assigns.requester_id
+      org_id = conn.assigns.org_id
+      user_id = conn.assigns.user_id
+
```
dexyk, line 24: @VeljkoMaksimovic let's log (info lvl) here with `requester_id, org_id, user_id`

VeljkoMaksimovic: 👍

lucaspin (approved):

PR #213, opened 2025-04-09
feat(guard): create soft delete for organizations

forestileao:
## 📝 Description
As a plan for making organizations restorable. We are implementing a soft-delete for them.
In another PR, a garbage collector for projects will be implemented in guard.

Main changes:
Organization GET operations only searches for orgs where deleted_at is null
Created Deleted At column to know when it was deleted for the GarbageCollector

## ✅ Checklist

### DIFF
--- a/github_hooks/db/migrate/20250409112220_add_deleted_at_for_organizations.rb
+++ b/github_hooks/db/migrate/20250409112220_add_deleted_at_for_organizations.rb
@@ -1,0 +1,6 @@
+class AddDeletedAtForOrganizations < ActiveRecord::Migration[5.1]
+  def change
+    add_column :organizations, :deleted_at, :datetime, null: true, default: nil
+    add_column :organizations, :deleted_by, :uuid, null: true, default: nil
+  end
+end
--- a/github_hooks/db/migrate/20250409112220_add_deleted_at_for_organizations.rb
+++ b/github_hooks/db/migrate/20250409112220_add_deleted_at_for_organizations.rb
@@ -1,6 +1,5 @@
 class AddDeletedAtForOrganizations < ActiveRecord::Migration[5.1]
   def change
     add_column :organizations, :deleted_at, :datetime, null: true, default: nil
-    add_column :organizations, :deleted_by, :uuid, null: true, default: nil
   end
 end
--- a/guard/lib/guard/store/organization.ex
+++ b/guard/lib/guard/store/organization.ex
@@ -8,6 +8,7 @@ defmodule Guard.Store.Organization do
   def get_by_id(id) when is_binary(id) and id != "" do
     Guard.FrontRepo.Organization
     |> where([o], o.id == ^id)
+    |> where_undeleted()
     |> Guard.FrontRepo.one()
     |> case do
       nil -> Util.ToTuple.error("Organization '#{id}' not found.", :not_found)
@@ -20,6 +21,7 @@ defmodule Guard.Store.Organization do
   def get_by_username(username) when is_binary(username) and username != "" do
     Guard.FrontRepo.Organization
     |> where([o], o.username == ^username)
+    |> where_undeleted()
     |> Guard.FrontRepo.one()
     |> case do
       nil -> Util.ToTuple.error("Organization '#{username}' not found.", :not_found)
@@ -30,6 +32,7 @@ defmodule Guard.Store.Organization do
   def exists?(org_id) do
     Guard.FrontRepo.Organization
     |> where([o], o.id == ^org_id)
+    |> where_undeleted()
     |> Guard.FrontRepo.exists?()
   end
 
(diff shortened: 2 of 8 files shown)


github_hooks/db/migrate/20250409112220_add_deleted_at_for_organizations.rb
```diff
@@ -0,0 +1,5 @@
+class AddDeletedAtForOrganizations < ActiveRecord::Migration[5.1]
+  def change
+    add_column :organizations, :deleted_at, :datetime, null: true, default: nil
```
VeljkoMaksimovic, line 3: in projects we added "deleted_by". That seems like a good idea, why not add it here as well?

guard/lib/guard/store/organization.ex
```diff
@@ -296,12 +301,28 @@ defmodule Guard.Store.Organization do
...
   Returns {:ok, organization} if successful, or {:error, changeset} if not.
   """
-  @spec destroy(Guard.FrontRepo.Organization.t()) ::
+  @spec soft_destroy(Guard.FrontRepo.Organization.t()) ::
           {:ok, Guard.FrontRepo.Organization.t()} | {:error, Ecto.Changeset.t()}
-  def destroy(%Guard.FrontRepo.Organization{} = organization) do
+  def soft_destroy(%Guard.FrontRepo.Organization{} = organization) do
+    organization
+    |> Guard.FrontRepo.Organization.changeset(%{deleted_at: DateTime.utc_now()})
+    |> Guard.FrontRepo.update()
```
VeljkoMaksimovic: I suggest we publish the soft_deleted event here, not in the controller

guard/lib/guard/store/organization.ex
```diff
@@ -296,12 +301,28 @@ defmodule Guard.Store.Organization do
...
+  end
+
+  @doc """
+  Deletes an organization completely from the database.
+  Returns {:ok, organization} if successful, or {:error, changeset} if not.
+  """
+  @spec hard_destroy(Guard.FrontRepo.Organization.t()) ::
+          {:ok, Guard.FrontRepo.Organization.t()} | {:error, Ecto.Changeset.t()}
+  def hard_destroy(%Guard.FrontRepo.Organization{} = organization) do
     Guard.FrontRepo.delete(organization)
```
VeljkoMaksimovic: Same for emitting the deleted event

github_hooks/db/migrate/20250409112220_add_deleted_at_for_organizations.rb
forestileao: We added that for projects because we validate the requester_id for deleting Schedulers for HARD_DELETION. So we would need that information persisted for future deletion.

In this case, we don't track the requester ID like we have for projects. To add it we would need to change the API protofiles as well as all the calls to it.

guard/lib/guard/store/organization.ex
```diff
@@ -296,12 +301,37 @@ defmodule Guard.Store.Organization do
...
           {:ok, Guard.FrontRepo.Organization.t()} | {:error, Ecto.Changeset.t()}
-  def destroy(%Guard.FrontRepo.Organization{} = organization) do
-    Guard.FrontRepo.delete(organization)
+  def soft_destroy(%Guard.FrontRepo.Organization{} = organization) do
+    result =
+      organization
+      |> Guard.FrontRepo.Organization.changeset(%{deleted_at: DateTime.utc_now()})
+      |> Guard.FrontRepo.update()
+
+    Guard.Events.OrganizationDeleted.publish(organization.id, soft_delete: true)
```
lucaspin: If the update fails, we don't want to send the event, right?

guard/lib/guard/store/organization.ex
```diff
@@ -296,12 +301,37 @@ defmodule Guard.Store.Organization do
...
+    result
+  end
+
+  @doc """
+  Deletes an organization completely from the database.
+  Returns {:ok, organization} if successful, or {:error, changeset} if not.
+  """
+  @spec hard_destroy(Guard.FrontRepo.Organization.t()) ::
+          {:ok, Guard.FrontRepo.Organization.t()} | {:error, Ecto.Changeset.t()}
+  def hard_destroy(%Guard.FrontRepo.Organization{} = organization) do
```
lucaspin, line 331: When and from where are we calling this?

guard/lib/guard/store/organization.ex
pedroforesti: Correct! Nice catch

guard/lib/guard/store/organization.ex
pedroforesti: we are keeping this for using it in the garbage collector worker. It is not implemented yet, but it will be used for sure

VeljkoMaksimovic (approved):

lucaspin (approved):

PR #215, opened 2025-04-10
feat(projecthub): create project cleaner

forestileao:
## 📝 Description
This creates a Project Cleaner for projects that were deleted after 30 days.

## ✅ Checklist


projecthub/lib/projecthub/workers/project_cleaner.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Projecthub.Workers.ProjectCleaner do
+  use Quantum, otp_app: :projecthub
+
+  alias Projecthub.Models.Project
+
+  require Logger
+
+  def process do
+    Project.find_candidates_for_hard_destroy()
```
hamir-suspect: We should lock the rows selected here maybe (in case 2 of these workers are started at least one would crash while mapping over projects to delete)

projecthub/lib/projecthub/models/project.ex
```diff
@@ -320,6 +320,20 @@ defmodule Projecthub.Models.Project do
     {:ok, nil}
   end
 
+  def find_candidates_for_hard_destroy() do
```
hamir-suspect, line 322: maybe this should be chunked/paginated as well

forestileao: It was before, but since we are not going to have a bunch of projects being deleted in a month. And also because of this potential problem:
https://github.com/semaphoreio/semaphore/pull/216#discussion_r2039484692

projecthub/lib/projecthub/workers/project_cleaner.ex
```diff
@@ -0,0 +1,17 @@
...
+    |> Enum.each(fn project ->
+      Logger.info("Hard destroying project #{project.id}")
+      Project.hard_destroy(project, project.deleted_by)
```
hamir-suspect: Please add metrics and error logging

hamir-suspect (changes requested):

hamir-suspect (approved):

PR #216, opened 2025-04-11
feat(guard): create organization cleaner

forestileao:
## 📝 Description
Creation of an organization cleaner that runs in background and searches for deleted organizations candidates to hard deletion every midnight.

## ✅ Checklist


guard/lib/guard/application.ex
```diff
@@ -70,13 +70,15 @@ defmodule Guard.Application do
     rabbit = System.get_env("RABBIT_CONSUMER") || "true"
     id = System.get_env("ID_API") || "false"
     instance_config_api = System.get_env("INSTANCE_CONFIG_API") || "false"
+    organization_cleaner = System.get_env("ORGANIZATION_CLEANER") || "false"
```
VeljkoMaksimovic: Nit pick, I would rename this to `enable_organization_cleaner?` or `start_organization_cleaner?`

guard/lib/guard/organization_cleaner.ex
```diff
@@ -0,0 +1,23 @@
...
+  alias Guard.Store.Organization
+
+  require Logger
+
+  @chunk_size 10
+
+  def process do
+    Organization.find_candidates_for_hard_destroy()
+    |> Stream.chunk_every(@chunk_size)
+    |> Stream.map(fn projects ->
```
VeljkoMaksimovic: Why is this variable named projects? Shouldn't it be `orgs` or `organizations`?

guard/lib/guard/organization_cleaner.ex
```diff
@@ -0,0 +1,23 @@
...
+
+  alias Guard.Store.Organization
+
+  require Logger
+
+  @chunk_size 10
+
+  def process do
+    Organization.find_candidates_for_hard_destroy()
+    |> Stream.chunk_every(@chunk_size)
```
VeljkoMaksimovic: One more question: What was the logic behind using Stream here instead of a simple Enum pipeline?

guard/lib/guard/organization_cleaner.ex
forestileao: Sorry, copy and paste from project cleaner

guard/lib/guard/organization_cleaner.ex
forestileao: It was just to run things more efficiently in case someones deletes a lot of things in a month. But I just noticed that my concurrency model was not making sense because of that internal Enum.map.

Check now, I added Task.async_stream

lucaspin (dismissed):

guard/lib/guard/store/organization.ex
```diff
@@ -341,6 +341,18 @@ defmodule Guard.Store.Organization do
     end
   end
 
+  @spec find_candidates_for_hard_destroy() :: [Guard.FrontRepo.Organization.t()]
+  def find_candidates_for_hard_destroy do
+    thirty_days_ago =
+      DateTime.utc_now()
+      |> DateTime.add(-30 * 24 * 60 * 60)
```
lucaspin: This should be probably be configurable

guard/lib/guard/organization_cleaner.ex
```diff
@@ -0,0 +1,29 @@
...
+    |> Stream.each(fn orgs ->
+      orgs
+      |> Task.async_stream(
+        fn org ->
+          Logger.info("Hard destroying organization #{org.id}")
+          Organization.hard_destroy(org)
+        end,
+        max_concurrency: @chunk_size,
```
VeljkoMaksimovic: This line is redundant. Each chunk has at max @chunk_size elements, Task.async_stream() has no way to start more than @chunk_size tasks, even without this specification

guard/lib/guard/store/organization.ex
forestileao: Updated!

guard/lib/guard/organization_cleaner.ex
VeljkoMaksimovic: I still dont like this approach, here are my 2 cents:

1) This will now start 10 parallel tasks all trying to modify something in the database, which will probably take up more than 10 DB connections from the DB pool. Looking at the configuration for this service (guard-consumer), the total size of pool is 10, so this would eat up all of the connections.
2) Hard destroy operation is basically one db update (as fast as it can get :D), so starting 10 separate tasks and running this operation in parallel probably does not provide any speedup compared to running it one by one


If you want to use Stream here, I suggest running Stream.flat_map and then Stream.each inside, IMO this should be a simple Enum.map() pipeline. The only benefit of stream (as far as I now), is not loading all of the data into the memory at once. Since all of the data in this case will be just a few organizations, loading it all into the memory does not seem to be a problem

forestileao: You are correct, this could cause issues because of the database connections.

VeljkoMaksimovic (dismissed):

forestileao:
@lucaspin @VeljkoMaksimovic 
Added some metrics and error logging. Also fixed its initialization in the Application, the worker reference was wrong
image

lucaspin (approved):

PR #228, opened 2025-04-16
feat(guard): add organization restore service

forestileao:
## 📝 Description

- Added organization restore service
- Add soft_delete param filters in describe, describe_many and list
- Added tests for the new service and for the changed ones

## ✅ Checklist

### DIFF
--- a/guard/lib/guard/store/organization.ex
+++ b/guard/lib/guard/store/organization.ex
@@ -71,9 +71,11 @@ defmodule Guard.Store.Organization do
 
   @spec list(map(), map()) :: {[FrontRepo.Organization.t()], String.t() | nil}
   def list(params, keyset_params) do
+    soft_deleted = Map.get(params, :soft_deleted, false)
+
     query =
       Guard.FrontRepo.Organization
-      |> filter_deleted_query(params.soft_deleted)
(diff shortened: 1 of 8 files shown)


guard/lib/guard/store/organization.ex
```diff
@@ -355,7 +375,11 @@ defmodule Guard.Store.Organization do
     |> Guard.FrontRepo.all()
   end
 
-  defp where_undeleted(query) do
-    query |> where([o], is_nil(o.deleted_at))
+  defp filter_deleted_query(query, soft_deleted) do
```
lucaspin: nitpick: I think Elixir pattern matching makes these types of functions easier to understand than an if statement:

```elixir
defp filter_deleted_query(query, false), do: query |> where([o], is_nil(o.deleted_at))
defp filter_deleted_query(query, true), do: query |> where([o], not is_nil(o.deleted_at))
```

lucaspin (approved):

guard/lib/guard/store/organization.ex
pedroforesti: Refactored!

lucaspin (approved):

PR #229, opened 2025-04-16
chore(e2e): introduce user login case and git integrations page tests

hamir-suspect:
## 📝 Description
Allows for running tests from local dev env.
Adds test case module that starts the session from a logged in user
Ads UI tests for git integrations page

## ✅ Checklist


lucaspin:
@hamir-suspect I don't see a successful CI run for these new tests. Am I missing something?

hamir-suspect:
@lucaspin the e2e tests have to be run manually as a promotion, do you think we should have automatic promotions when there is a change in `e2e` path?

lucaspin:
@hamir-suspect that would be nice, but I don't think you have to do that as part of this PR

e2e/test/e2e/ui/git_integrations_test.exs
```diff
@@ -0,0 +1,164 @@
...
+
+    test "has two main sections: Integrations and Connect new", %{session: session} do
+      session
+      |> assert_has(
+        Wallaby.Query.css(".bb.b--black-075.w-100-l.mb4.br3.shadow-3.bg-white",
+          count: 2
+        )
+      )
+      |> assert_has(
+        Wallaby.Query.css(".bb.bw1.b--black-075.br3.br--top .flex.items-center .b",
```
lucaspin: Can't we use IDs to find these elements? If we assert on these CSS classes, every time we change something on the CSS, we'll have to update this test. We could also use a more distinctive class like ".integrations-list" and "interrations-connect-new"  instead of these generic style-related ones.

e2e/docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+version: "3"
+
+services:
+  e2e-tests:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    platform: linux/arm64 # Use ARM64 platform for Apple Silicon Macs
```
lucaspin: I assume this is needed because of the chromium driver binary, but can we parameterize it, so we can run on linux/amd64 as well? Maybe we can have a `PLATFORM` in the Makefile which defaults to `linux/arm64`, and is passed here, so you can run `make test.ui.docker PLATFORM=linux/amd64`

lucaspin (approved):

lucaspin (changes requested):
I approved before because I forgot about my other comment about the use of CSS classes in the wallaby tests 😅

e2e/test/support/ui_test_case.ex
```diff
@@ -0,0 +1,81 @@
...
+        |> (fn s ->
+            # Verify login form exists
+            has?(s, Wallaby.Query.css("#kc-form-login"))
+            s
+          end).()
+        |> fill_in(Wallaby.Query.text_field("username"), with: root_email)
+        |> fill_in(Wallaby.Query.text_field("password"), with: root_password)
+        |> click(Wallaby.Query.css("#kc-login"))
+
+      assert current_url(logged_in_session) == "https://#{organization}.#{base_domain}/get_started/"
```
skipi: I'm not sure if we should make this assertion here. I can see a scenario where this might cause us some pain(testing redirects, onboarding already completed). 🤔

front/assets/js/git_integration/components/card.tsx
```diff
@@ -22,7 +22,7 @@ interface CardProps {
 
 export const Card = (props: CardProps) => {
   return (
-    <div className={`ph3 pv2 mv2 ${props.lastItem ? `` : `bb b--black-075`}`}>
+    <div data-testid="integration-card" className={`ph3 pv2 mv2 ${props.lastItem ? `` : `bb b--black-075`}`}>
```
skipi, line 25: I don't like the name of the `data-testid` attribute. WDYT about changing it to something like `data-element-id`? 🤔

hamir-suspect: the `data-testid` seems to be the [standard practice](https://www.educative.io/answers/what-is-the-data-testid-attribute-in-testing) for assuring stable DOM selectors in tests.

e2e/test/support/ui_test_case.ex
hamir-suspect: This is addressed in following PRs in this e2e queue, backporting the changes here.

skipi (approved):

lucaspin (approved):

PR #231, opened 2025-04-16
Front/fix group creation

VeljkoMaksimovic:
## 📝 Description

Several improvements to the groups UI/service:

- On the people page, you can no longer click on the group (like you can on the regular user, that takes you to the profile page)
- Implement destroy procedure in the group backed service (This also includes some changes to the groups table, hence the migrations in the guard and rbac repos)
- If you go to the people page (on org or project), the 'x' next to the subject retracts the role, but now the 'x' next to the group destroys the group (If you are on the organization people page)

https://github.com/renderedtext/tasks/issues/7771

## ✅ Checklist


front/assets/js/groups/group_management.js
```diff
@@ -266,5 +267,31 @@ export var GroupManagement = {
...
+  },
+
+  areMandatoryFieldsEmpty() {
+    const name = document.getElementById("name_input").value
+    const description = document.getElementById("description_input").value
+    return !name || !description
+  },
+
+  isAnyChangeMade() {
+    var isNameChanged, isDescriptionChanged
```
pedroforesti:
```suggestion
    let isNameChanged, isDescriptionChanged
```

front/lib/front_web/controllers/groups_controller.ex
```diff
@@ -141,4 +141,33 @@ defmodule FrontWeb.GroupsController do
 
     conn |> json(group_members)
   end
+
+  def destroy_group(conn, params) do
+    org_id = conn.assigns.organization_id
```
pedroforesti, line 146: Should we add groups feature flag verification for this org?

VeljkoMaksimovic: Good catch! I want to create a global plug for authorizing feature (checking if the feature flag is enabled). This change would mean updating a bunch of controllers, so I will do it in a separate PR

pedroforesti (approved):

forestileao (approved):

PR #261, opened 2025-04-28
feat(deliveryhub): implement RabbitMQ events

forestileao:
### Changes:
- Created RabbitMQ events for:
  - stage created
  - event source created
  - stage event created (event enqueued on stage)
  - stage event approved
  - execution created
  - execution started
  - execution finished

Note: The routing key depends on pointing directly to each parent resource Id to make the available filtering in the UI

## ✅ Checklist

### DIFF
--- a/ee/delivery-hub/pkg/grpc/actions/messages/event_source_created.go
+++ b/ee/delivery-hub/pkg/grpc/actions/messages/event_source_created.go
@@ -16,6 +16,7 @@ const EventSourceCreatedRoutingKey = "event-source-created"
 func NewEventSourceCreatedMessage(eventSource *models.EventSource) EventSourceCreatedMessage {
 	return EventSourceCreatedMessage{
 		message: &pb.EventSourceCreated{
+			CanvasId:  eventSource.CanvasID.String(),
 			SourceId:  eventSource.ID.String(),
 			Timestamp: timestamppb.Now(),
 		},
--- a/ee/delivery-hub/pkg/grpc/actions/messages/event_source_created.go
+++ b/ee/delivery-hub/pkg/grpc/actions/messages/event_source_created.go
@@ -3,20 +3,25 @@ package messages
 import (
 	"github.com/semaphoreio/semaphore/delivery-hub/pkg/models"
 	pb "github.com/semaphoreio/semaphore/delivery-hub/pkg/protos/delivery"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 type EventSourceCreatedMessage struct {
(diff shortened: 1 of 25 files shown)


ee/delivery-hub/pkg/grpc/actions/messages/event_source_created.go
```diff
@@ -0,0 +1,35 @@
...
+	pb "github.com/semaphoreio/semaphore/delivery-hub/pkg/protos/delivery"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type EventSourceCreatedMessage struct {
+	message  *pb.EventSourceCreated
+	canvasId string
+}
+
+const EventSourceCreatedExchange = "DeliveryHub.EventSourceExchange"
```
lucaspin: What is the reason for using separate exchanges for event sources, executions, stage and stage events? The UI will need to create multiple queues/consumers to listen to all these exchanges, no?

pedroforesti: In this case, the exchange names are different, this is just to segregate by the resource, but the names are static. The major problem I see is that the routing key will change according to the parent resource ID and if we use the current implementation of Tackle, maybe it will need to create consumers dynamically anyway since all exchanges will be of type direct (hardcoded in Tackle).

pedroforesti: There are 2 main approaches we need to discuss:
- The front will use dynamic Consumers changing the routing key (current implementation)
- The front will receive all events and in some way send them to the correct Canvas Client UI based on the message body

ee/delivery-hub/pkg/grpc/actions/messages/utils.go
```diff
@@ -0,0 +1,33 @@
...
+
+	return tackle.PublishMessage(&tackle.PublishParams{
+		Body:       message,
+		AmqpURL:    amqpURL,
+		RoutingKey: routingKey,
+		Exchange:   exchange,
+	})
+}
+
+func toJSON(m interface{}) []byte {
```
lucaspin: You wouldn't encode it as JSON. We need to encode it as a protobuf message. We can use `proto.Marshal`, like [here](https://github.com/semaphoreio/semaphore/blob/feature/delivery/ee/delivery-hub/pkg/workers/pipeline_done_consumer_test.go#L49)

forestileao: I had used another example before. Now we should be good.

lucaspin (approved):

PR #272, opened 2025-05-06
feat(delivery): time window condition

lucaspin:
### Database changes

- New conditions JSON column on stages table, which will hold all the conditions that must pass for the stage.
- Stage events table has new column `state_reason`, which so far can be `time-window` or `approval`. The state  of the stage will be `waiting` for both of those cases.

### API changes

Changes from https://github.com/renderedtext/internal_api/pull/843.

### Worker updates

- New time window worker, which works on stage events with state = waiting and state_reason = time-window. Controlled by `START_TIME_WINDOW_WORKER` environment variable.
- New worker that consumes messages about stage event approvals, and moves the stage event back to the pending if the number of approvals has been reached. Controlled by `START_STAGE_EVENT_APPROVED_CONSUMER` environment variable.
- Pending stage events worker was updated to process time window conditions as well.

### DIFF
--- a/ee/delivery-hub/pkg/models/stage.go
+++ b/ee/delivery-hub/pkg/models/stage.go
@@ -10,19 +10,38 @@ import (
 )
 
 const (
-	RunTemplateTypeSemaphore = "semaphore"
+	RunTemplateTypeSemaphore     = "semaphore"
+	StageConditionTypeApproval   = "approval"
+	StageConditionTypeTimeWindow = "time-window"
 )
 
 type Stage struct {
-	ID               uuid.UUID `gorm:"type:uuid;primary_key;"`
-	OrganizationID   uuid.UUID
-	CanvasID         uuid.UUID
-	Name             string
-	CreatedAt        *time.Time
-	CreatedBy        uuid.UUID
-	ApprovalRequired bool
+	ID             uuid.UUID `gorm:"type:uuid;primary_key;"`
+	OrganizationID uuid.UUID
+	CanvasID       uuid.UUID
+	Name           string
+	CreatedAt      *time.Time
+	CreatedBy      uuid.UUID
 
+	Conditions  datatypes.JSONSlice[StageCondition]
 	RunTemplate datatypes.JSONType[RunTemplate]
+}
+
+type StageCondition struct {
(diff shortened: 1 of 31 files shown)


ee/delivery-hub/pkg/models/stage.go
```diff
@@ -10,21 +12,120 @@ import (
...
+
+	return &TimeWindowCondition{
+		Start:    start,
+		End:      end,
+		WeekDays: days,
+	}, nil
+}
+
+// Copied from Golang's time package
+var layout = "15:04"
```
pedroforesti: I know it is copied from the time package, but do we know why 15:04? If so, it would be nice to add the explanation.

ee/delivery-hub/pkg/workers/pending_stage_events_worker.go
```diff
@@ -85,21 +94,19 @@ func (w *PendingStageEventsWorker) ProcessEvent(stage *models.Stage, event *mode
...
+		if err != nil {
+			return err
 		}
 
-		logger.Infof("Event %s waiting for approval", event.ID)
-
-		return nil
+		if !proceed {
+			return nil
+		}
```
pedroforesti, line 107: I know it was already like this before, but why are we returning nil and not an error if we are not proceeding with the condition?

lucaspin: Because we check for an error in the line above, and we want to exit early from ProcessEvent(), since we already found one condition that is blocking this event.

```go
for _, condition := range stage.Conditions {
  proceed, err := w.checkCondition(logger, event, condition)

  // here, an actual error happened while trying
  // to evaluate the condition, so we return it
  if err != nil {
    return err
  }

  // here, no error happened, and we know that this condition didn't pass,
  // so we return nil to exit early from the ProcessEvent() function without
  // doing anything else for this event
  if !proceed {
    return nil
  }
}
```

ee/delivery-hub/pkg/models/stage.go
lucaspin: Done

forestileao (approved):

PR #273, opened 2025-05-06
feat(delivery-hub): add semaphore event source handling

forestileao:
## 📝 Description
- Add semaphore event source handling
- Added some todos for creating secret and notification
- Updated organization middleware

## ✅ Checklist

### DIFF
--- a/ee/delivery-hub/pkg/public/org_middleware.go
+++ b/ee/delivery-hub/pkg/public/org_middleware.go
@@ -1,8 +1,13 @@
 package public
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"fmt"
+	"io"
 	"net/http"
+	"strings"
 
 	"github.com/google/uuid"
 )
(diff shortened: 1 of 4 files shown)


ee/delivery-hub/pkg/public/org_middleware.go
```diff
@@ -1,37 +1,132 @@
...
 var orgIDKey contextKey = "org-id"
 
+type SourceType string
+
+const (
+	SourceTypeGithub    SourceType = "github"
+	SourceTypeSemaphore SourceType = "semaphore"
+)
+
 func OrganizationMiddleware(next http.Handler) http.Handler {
```
lucaspin: We should always get the organization ID from the `x-semaphore-*` headers

pedroforesti: But the notifications don't send this header, correct?
Should I change this implementation as well?

https://github.com/semaphoreio/semaphore/blob/main/notifications/lib/notifications/workers/webhook.ex#L47-L59

lucaspin: No. The notification service doesn't send it, but this request should go through the auth service and that service will validate the organization, and will include the `x-semaphore-*` headers if the organization is valid.

lucaspin (approved):

PR #280, opened 2025-05-08
feat(delivery-hub): Create "Update Stage" Service

forestileao:
## 📝 Description
- Create "Update Stage" Service
- Create rabbit mq message to send event that stage was updated

## ✅ Checklist


forestileao:
@lucaspin Probably we still need to add a field "name" in UpdateStageRequest, so we don't need to always depend on the ID.
But the feature looks like it it is working using the cli: https://github.com/semaphoreci/cli/pull/236


image

PR #283, opened 2025-05-09
Mk/generic git support

skipi:
## ✅ Checklist


hooks_processor/lib/hooks_processor/clients/repository_client.ex
```diff
@@ -12,6 +19,93 @@ defmodule HooksProcessor.Clients.RepositoryClient do
   @wormhole_timeout 6_000
   @grpc_timeout 5_000
 
+  # RegenerateWebhook
+
+  def regenerate_webhook(repository_id) do
```
hamir-suspect: 🔪 if it's not used

hooks_receiver/lib/hooks_receiver/repository_client.ex
```diff
@@ -32,9 +38,53 @@ defmodule HooksReceiver.RepositoryClient do
...
+          |> Map.get(:webhook)
+          |> ToTuple.ok()
+
+        {:error, error} ->
+          error
+      end
+    end)
+  end
+
+  def regenerate_webhook(repository_id) do
```
hamir-suspect: We have it here as well

hamir-suspect (changes requested):

front/lib/front/models/commit_job.ex
```diff
@@ -146,18 +178,65 @@ defmodule Front.Models.CommitJob do
...
   end
 
   defp add_commands_for_modified_files(params) do
     Enum.reduce(params.changes, [], fn change, acc ->
       if change.action != Action.value(:DELETE_FILE) do
-        acc ++ ["mv ../.changed_files/#{change.file.path} ./#{change.file.path}"]
+        dir_path = Path.dirname(change.file.path)
+
+        acc ++
+          ["mkdir -p ./#{dir_path}"] ++
```
DamjanBecirovic, line 234: This will fail if the directory already exists (which is the case when editing existing files from the workflow editor) which would also fail the whole commit job.

skipi: 🤔 I think I've checked this - can't reproduce the error ⏬ 

```bash
docker run -it alpine:latest /bin/ash
/ # ls
bin    dev    etc    home   lib    media  mnt    opt    proc   root   run    sbin   srv    sys    tmp    usr    var
/ # cd tmp
/tmp # ls
/tmp # mkdir -p foo/bar/baz
/tmp # tree
.
└── foo
    └── bar
        └── baz

3 directories, 0 files
/tmp # mkdir -p foo/bar/baz
/tmp # echo $?
0
/tmp # mkdir -p foo/bar/baz
/tmp # mkdir -p foo/bar/baz
/tmp # mkdir -p foo/bar/baz
/tmp # echo $?
0
/tmp # mkdir -p foo/bar/az
/tmp # echo $?
0
/tmp # mkdir -p foo/bar/321/fdsa
/tmp # mkdir -p foo/bar/321/cxzcxz/az
/tmp # tree
.
└── foo
    └── bar
        ├── 321
        │   ├── cxzcxz
        │   │   └── az
        │   └── fdsa
        ├── az
        └── baz

8 directories, 0 files
/tmp #
```

front/lib/front/models/commit_job.ex
```diff
@@ -146,18 +178,65 @@ defmodule Front.Models.CommitJob do
...
+    else
+      [
+        # Fetch remote branches
+        "git fetch origin #{target} || true",
+        # Check if branch exists on remote
+        "if git show-ref --verify --quiet refs/remotes/origin/#{target}; then",
+        "  git checkout -B #{target} origin/#{target}",
+        "else",
+        "  git checkout -B #{target} #{initial}",
+        "fi"
```
DamjanBecirovic: I'm not 100%sure how this multi-line if-else will work on Zebra, but I assume it would be interpreted as multiple one-line commands. Is this working as expected when tested manually?

skipi: I think I can do a one-liner from this

front/lib/front/models/commit_job.ex
DamjanBecirovic: I tried it without the `-p` flag 🤦‍♂️

front/lib/front_web/controllers/project_controller.ex
```diff
@@ -677,21 +678,38 @@ defmodule FrontWeb.ProjectController do
...
+          url
+        end
+      end
+
+    conn
+    |> put_status(200)
+    |> json(%{
+      workflow_url: workflow_url,
+      artifact_url: artifact_url
+    })
```
DamjanBecirovic, line 712: There is already this [check_commit_job](https://github.com/semaphoreio/semaphore/blob/main/front/lib/front_web/controllers/project_controller.ex#L538) endpoint for this check, why not use that one?

skipi: The only reason is that we're already using this endpoint in the project onboarding - it made the integration with the UI a little bit easier.

hamir-suspect (approved):

DamjanBecirovic (approved):

hamir-suspect (approved):

DamjanBecirovic (approved):

forestileao (approved):

PR #295, opened 2025-05-13
feat(delivery-hub): add header filter type

forestileao:
## 📝 Description
- Add header Filter type for conditions
- Header names should be case sensitive
- Add Header function to expression parser because of dashed header names

## ✅ Checklist

### DIFF
--- a/ee/delivery-hub/db/migrations/20250512212918_add-stage-event-headers-column.down.sql
+++ b/ee/delivery-hub/db/migrations/20250512212918_add-stage-event-headers-column.down.sql
@@ -1,0 +1,5 @@
+begin;
+
+ALTER_TABLE events DROP COLUMN headers;
+
+commit;
--- a/ee/delivery-hub/db/migrations/20250512212918_add-stage-event-headers-column.down.sql
+++ b/ee/delivery-hub/db/migrations/20250512212918_add-stage-event-headers-column.down.sql
@@ -1,5 +1,0 @@
-begin;
-
-ALTER_TABLE events DROP COLUMN headers;
-
-commit;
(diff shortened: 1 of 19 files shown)


ee/delivery-hub/db/migrations/20250512212918_add-stage-event-headers-column.down.sql
```diff
@@ -0,0 +1,5 @@
+begin;
```
lucaspin: We don't do rollbacks for database like this, so this is not needed

ee/delivery-hub/db/migrations/20250512212918_add-stage-event-headers-column.up.sql
```diff
@@ -0,0 +1,5 @@
+begin;
+
+ALTER TABLE events ADD COLUMN headers jsonb NOT NULL DEFAULT '{}';
```
lucaspin, line 3: Why not have the default be NULL?

ee/delivery-hub/pkg/models/event.go
```diff
@@ -231,3 +253,58 @@ func FindLastEventBySourceID(sourceID uuid.UUID) (map[string]any, error) {
...
+// caseSensitive: whether the expression should be case sensitive.
+func CompileBooleanExpression(variables map[string]any, expression string, caseSensitive bool) (*vm.Program, error) {
+	options := []expr.Option{
+		expr.Env(variables),
+		expr.AsBool(),
+		expr.WithContext("ctx"),
+		expr.Timezone(time.UTC.String()),
+	}
+
+	if !caseSensitive {
```
lucaspin: `caseSensitive` makes it seem like it also applies to the data filter, which is not true. It would be probably better to pass the filter type here directly.

ee/delivery-hub/db/migrations/20250512212918_add-stage-event-headers-column.up.sql
forestileao: Because all requests contain headers, even if they are empty. This is to avoid nil errors when querying events

Issue #297, opened 2025-05-13
[DOCS] Document how to use reports dashboard


TomFern:
## Documentation Location
Document how to use reports dashboard

## Type of Documentation Issue

- [X] Missing documentation
- [ ] Incorrect information
- [ ] Unclear explanation
- [ ] Outdated content
- [ ] Broken link
- [ ] Code example needs update
- [ ] Other (please describe)

## Current State
No documentation exists now

## Requested Changes
Show how to use and demostrate a few examples. Also update the test-report command reference

## Affects Version

- [X] Cloud Semaphore
- [X] Community Edition (specify version):

TomFern:
Merged

PR #313, opened 2025-05-20
docs: document how to use Markdown Reports in Cloud and CE

TomFern:
## 📝 Description
Document how to use Markdown Reports in Cloud and CE

See https://github.com/semaphoreio/semaphore/issues/297

## ✅ Checklist


TomFern:
@skipi could you take a look at these new pages for the Markdown Reports? The CE and Cloud versions are almost the same, the only difference is that I removed the resource-metrics parts in CE.

docs/docs/using-semaphore/tests/markdown-reports.md
```diff
@@ -0,0 +1,322 @@
...
+
+![Gantt commands](./img/timeline.jpg)
+
+Obtain resource utilization metrics.
+
+![Utilization metrics](./img/resource-metrics.jpg)
+
+You can even use images in your reports.
+
+![Embedded Image](./img/markdown-reports-dependencies-example.jpg)
```
AleksandarCole, line 41: @TomFern Awesome page 🙌 

I only found this part confusing because this image looks a lot like it's `mermaid.js` chart and not just an example of a random image you can embed.

I would consider: 
1. Using some other image, because that one looks a lot like it's actually markdown and not PNG. Maybe a cat image 🐱 Do kids these days still use those as a placeholder images or it's overused trope?
2. Would be cool if we had an actual example of generating this with dependency cruiser in markdown, or at least as an image that gets uploaded to report.

PR #321, opened 2025-05-21
fix(front): do not make job line commands sticky if they are too big

forestileao:
## 📝 Description

If a job line command is too big and sticky options are enabled in job page preferences, the job line command might hide all the content of a command output.
To avoid this, let's make a limitation for a command to not be sticky if its height is greater than 600px (our job line page is entirely 800px).

## ✅ Checklist


forestileao:
> This feels like a hack solution. Maybe instead we can collapse the command to one line and expand it when the whole command block is shown?

But the problem would persist if we expand, since the sticky option is what is causing the issue, right?
The problem happens only when the command block is expanded (with the output being shown) and the command that generates the output is sticky.

forestileao:
I mean, I don't think it's possible to keep this the job line command sticky if we have the whole logs container with height of 800px

pedroforesti:
Unless we add a "command line is too large" and a link to see the raw log. And limiting the max-height to 600px.
But only for the commands, not for the output.

@AleksandarCole WDYT?

radwo:
do you have an example of a given log with that issue?

radwo:
CleanShot 2025-05-21 at 21 12 16@2x

I was thinking of something like this, when the line is to long you can toggle it to show the whole content or be able to see what is below.

forestileao:
I see, to have the expansion even with the whole log open. That's a good idea. Looking forward to it. Just verifying if this is the definitive decision, and I will update the PR.

forestileao:
@radwo Just adding an example of logs here:

https://github.com/user-attachments/assets/86a43018-db20-47f4-a8fb-4c462203447c

skipi (approved):

radwo:
@forestileao can you put short video of how this is working now?

forestileao:
@radwo here it is:

https://github.com/user-attachments/assets/07e8071e-d20e-4180-be0c-c741cb249ea2

hamir-suspect (approved):
Looks ok to me 😄

radwo (approved):
Looks good 👍

PR #323, opened 2025-05-22
feat(v1alpha api): Allow passing parameters in Run Workflow API call

DamjanBecirovic:
## 📝 Description

Changes the Run workflow API implementation to allow passing parameters and improve performance.

Previously, the request was synchronously waiting for repo-proxy to create a hook by retrieving data from the git provider and after that to schedule the workflow on the plumber. 
This had performance issues and it did not support passing parameters to be used in the pipeline.

The new implementation goes directly to the plumber which only saves the request in the database before responding and all hook processing is done asynchronously. 
This should improve performance and also the plumber supports passing parameters out-of-the-box.

## ✅ Checklist
- [x] This change requires a documentation update


plumber/ppl/lib/ppl/workflow/workflow_api_server.ex
```diff
@@ -318,9 +318,8 @@ defmodule Plumber.WorkflowAPI.Server do
     end)
   end
 
-  defp params_contain_scheduler_task_id?(params) do
-    params |> Map.get("scheduler_task_id", "") |> String.length() > 0
-  end
+  defp start_in_conceived_state?(%{"scheduler_task_id" => val}) when is_binary(val) and val != "", do: true
+  defp start_in_conceived_state?(%{"start_in_conceived_state" => val}), do: val
```
dexyk, line 322: @DamjanBecirovic should we have a third clause returning `false`?

DamjanBecirovic:
There is an issue with starting workflows on tags and PRs with this approach, thanks to @dexyk for finding that out.
I'll have to add some code in hooks-processor and repository-hub to handle those cases.
I'm converting this to a draft PR and will re-request the reviews once I fix that part.

plumber/ppl/lib/ppl/workflow/workflow_api_server.ex
DamjanBecirovic: The `start_in_conceived_state` flag comes from a gRPC request, so it guarantees that it is always present in the request and false by default. 
Due to that, the line 322 will always evaluate to false unless the flag is explicitly set to true in request.

plumber/ppl/lib/ppl/repo_proxy_client.ex
```diff
@@ -95,14 +100,16 @@ defmodule Ppl.RepoProxyClient do
...
 
   defp git_reference_from_ppl_req(ppl_req) do
-    branch_name = ppl_req.request_args |> Map.get("branch_name", "")
+    git_ref = ppl_req.request_args |> Map.get("git_reference", "")
 
-    if branch_name == "" do
-      raise "Provided an empty branch_name"
+    if git_ref == "" do
+      branch_name = ppl_req.request_args |> Map.get("branch_name", "")
+      "refs/heads/#{branch_name}"
```
dexyk, line 107: @DamjanBecirovic should we raise exception in case `branch_name` is empty

DamjanBecirovic: The `branch_name` can not be empty since we have [a check](https://github.com/semaphoreio/semaphore/blob/main/plumber/ppl/lib/ppl/ppl_requests/model/ppl_requests.ex#L249) for that in Ecto changeset applied when we are creating a pipeline. All schedule requests that do not have branch_name will be rejjected due to that.

dexyk (approved):

Issue #337, opened 2025-05-27
Enhancing Contribution and Developer Guidelines


AleksandarCole:
### Discussed in https://github.com/orgs/semaphoreio/discussions/94

<div type='discussions-op-text'>

<sup>Originally posted by **AleksandarCole** March  5, 2025</sup>

## Problem Statement
Our current contribution guidelines are very basic and lack the detailed information developers need to effectively contribute to the project. We need comprehensive documentation that covers all aspects of the development process, coding standards, and architectural decisions.

## Current State
- Basic contribution guidelines that don't provide enough direction
- Missing developer-focused documentation
- No clear coding standards or best practices documentation
- Lack of UI style guide for frontend development
- No Architecture Decision Records (ADRs) to explain key technical decisions

## Desired Outcome
Comprehensive documentation that includes:
- Detailed contribution guidelines
- Complete developer guidelines covering the entire development process
- Clear standards for code, commits, branches, and PRs
- UI style guide for consistent design (figure out how to utilize the existing `app-design` repo)
- Architecture documentation including ADRs

## Next Steps
Logging here the areas that need to be documented/covered in order of importance as sub-tasks. @VeljkoMaksimovic feel free to extract as separate tasks if it's easier to track:

- [ ] Core contribution guide (overall - find and follow existing standards) > @VeljkoMaksimovic 
  - From fork to PR (cloning, naming, basic code style guides, etc)
  - Submission/Review process (requirements for merge, what happens after submission)
  - Communication > where to ask questions/report bugs
- [ ] Code dev. documentation > @VeljkoMaksimovic 
  - How to run in local
  - How to test it properly
  - Architecture overview and key concepts
- [ ] Review and update security policy as needed > @AleksandarCole 
- [ ] Good first issue  > @AleksandarCole 
  - Pick and transfer tasks from internal tickets to public
  - Label and categorize the "Good first issue" ones
  - Update tasks to have detailed explanations for beginner friendly contribution
- [ ] Bug bounty > @AleksandarCole 
  - Document bug bounty program
  - Kick off few bounties on platform
  - Work /w other teams to spread the word
</div>

PR #356, opened 2025-06-05
Developer experience guidelines

skipi:
## 📝 Description
Part of https://github.com/semaphoreio/semaphore/issues/337

## ✅ Checklist


TomFern (approved):

radwo (commented):
I have one note here, dev containers are not limited only to VS Code. What do you think of updating the info that dev containers are supported, and only mention that you can use VS Code to run them? Currently it sounds as you should use it because there is no other way to run containers.

skipi:
> I have one note here, dev containers are not limited only to VS Code. What do you think of updating the info that dev containers are supported, and only mention that you can use VS Code to run them? Currently it sounds as you should use it because there is no other way to run containers.

Makes perfect sense. I'll update the guidelines.

TomFern (approved):

PR #358, opened 2025-06-05
Dev/consolidate services

VeljkoMaksimovic:
## 📝 Description

When running services using skaffold dev, consolidate different microservices into singular deployment
https://github.com/renderedtext/project-tasks/issues/2417

## ✅ Checklist


radwo (commented):
CleanShot 2025-06-05 at 17 51 02@2x

There is an inconsistency in naming here: on one hand, you have `minimalDeployment`, and on the other, `local-dev` and `LOCAL_DEV_ENV`. It would be better to keep the naming consistent to avoid confusion later.

Also, I think `minimalDeployment` is a better choice than `local-dev`, since someone might want to run this on a small production server with minimal traffic.

hamir-suspect (approved):

skipi (approved):

dexyk (approved):

lucaspin (approved):

PR #389, opened 2025-06-13
chore: reduce volume of logs produced by the api

hamir-suspect:
## 📝 Description
Enables configuration of rbac-ee log level with env var in runtime
reduces the ammount of logs that we produce in self hosted hub api by logging less data and only api errors
## ✅ Checklist


self_hosted_hub/pkg/publicapi/server.go
```diff
@@ -591,7 +592,7 @@ func (s *Server) Serve(host string, port int) error {
 		WriteTimeout: 30 * time.Second,
 		IdleTimeout:  60 * time.Second,
 		Handler: http.TimeoutHandler(
-			handlers.LoggingHandler(os.Stdout, s.Router),
+			errorOnlyLoggingHandler(os.Stdout, s.Router),
```
lucaspin: Do we really need this? Not having logs about 2xx requests is problematic when troubleshooting issues with agent communication. Can we make only the `pkg/logging/logging.go` change first, and see the how that will affect the logs generated, and if not enough, we do this one?

lucaspin (approved):

PR #416, opened 2025-06-27
feat(rbac): enable okta login when creating integration

VeljkoMaksimovic:
## 📝 Description

https://github.com/renderedtext/tasks/issues/8169

## ✅ Checklist


ee/rbac/lib/rbac/okta/integrations.ex
```diff
@@ -32,12 +34,39 @@ defmodule Rbac.Okta.Integration do
              jit_provisioning_enabled: jit_provisioning_enabled,
              idempotency_token: idempotency_token
            ) do
+      add_okta_to_allowed_id_providers(org_id)
```
lucaspin, line 42: Shouldn't we handle the possible `{:error, e}` tuple here?

VeljkoMaksimovic: Good point, I've restructured this function to be within a db transaction

PR #466, opened 2025-08-06
Add service account feature

skipi:
## 📝 Description
Combines front/backend changes for service account functionality.

## ✅ Checklist


VeljkoMaksimovic (approved):

guard/lib/guard/front_repo/user.ex
```diff
@@ -229,4 +239,59 @@ defmodule Guard.FrontRepo.User do
...
+
+  @doc """
+  Generates a synthetic email for a service account.
+  """
+  def generate_service_account_email(service_account_name, organization_name) do
+    # Sanitize names to ensure valid email format
+    sanitized_sa_name = sanitize_email_part(service_account_name)
+    sanitized_org_name = sanitize_email_part(organization_name)
+
+    "#{sanitized_sa_name}@sa.#{sanitized_org_name}.semaphoreci.com"
```
hamir-suspect: @DamjanBecirovic thoughts on this?

DamjanBecirovic: The format looks good, but consider changing the `sa` to either `service-accounts` or `service_accounts`, as the `sa` abbreviation is not very common and may not be recognizable.

hamir-suspect: I also removed hardcoded "semaphoreci.com" and used `BASE_DOMAIN` env var.

DamjanBecirovic (approved):

hamir-suspect (dismissed):

VeljkoMaksimovic (dismissed):

VeljkoMaksimovic (approved):

PR #505, opened 2025-08-19
feat(front): display who was last to modify notification

VeljkoMaksimovic:
## 📝 Description

https://github.com/renderedtext/project-tasks/issues/2659

## ✅ Checklist

### DIFF
--- a/front/lib/front_web/controllers/notifications_controller.ex
+++ b/front/lib/front_web/controllers/notifications_controller.ex
@@ -307,7 +307,8 @@ defmodule FrontWeb.NotificationsController do
 
     map(notifications, fn n ->
       if n.metadata.creator_id != "" do
-        creator = find(creators, &(&1.id == n.metadata.creator_id)) || %{name: "deleted_user"}
+        default_username = Application.get_env(:front, :default_user_name)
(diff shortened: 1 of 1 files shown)


hamir-suspect (dismissed):

front/lib/front_web/controllers/notifications_controller.ex
```diff
@@ -298,6 +295,26 @@ defmodule FrontWeb.NotificationsController do
...
+    creators =
+      notifications
+      |> map(& &1.metadata.creator_id)
+      |> reject(&(&1 == ""))
+      |> uniq()
+      |> User.find_many()
+
+    map(notifications, fn n ->
+      if n.metadata.creator_id != "" do
+        creator = find(creators, &(&1.id == n.metadata.creator_id)) || %{name: "deleted_user"}
```
DamjanBecirovic: For deleted users, we should use [the same string](https://github.com/semaphoreio/semaphore/blob/main/front/lib/front_web/controllers/schedulers_controller.ex#L656) as in other places.

VeljkoMaksimovic: Didn't know we have that :D Added it

DamjanBecirovic (approved):

hamir-suspect (approved):

PR #570, opened 2025-09-12
fix(self-hosted-hub): fix agent cleanup logic and handle job failures during cleanup

dexyk:
## 📝 Description
Improves the agent cleanup process in the self-hosted hub. It ensures that when agents are removed, either because they are stuck or disconnected, any jobs assigned to those agents are properly finalized and marked as failed, with the corresponding messages published to RabbitMQ.

Related [task](https://github.com/renderedtext/tasks/issues/8579).

## ✅ Checklist

### DIFF
--- a/self_hosted_hub/pkg/workers/disconnectedcleaner/worker.go
+++ b/self_hosted_hub/pkg/workers/disconnectedcleaner/worker.go
@@ -1,8 +1,10 @@
 package disconnectedcleaner
 
 import (
+	"context"
 	"time"
 
+	"github.com/semaphoreio/semaphore/self_hosted_hub/pkg/amqp"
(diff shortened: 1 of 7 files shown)


self_hosted_hub/pkg/workers/disconnectedcleaner/worker.go
lucaspin, line 1: Not sure we should do the same for this worker. This worker handles agents in disconnected state, and agents only end up in that state when the agent itself disconnects, which only happens after they report a job_finished, so we shouldn't have the same problem for these.

Also, we don't release agents that will be shut down (to not have then reassigned to new jobs), so the assigned_job_id might still be set when this worker runs. See: https://github.com/semaphoreio/semaphore/blob/main/self_hosted_hub/pkg/agentsync/protocol.go#L227

PR #581, opened 2025-09-17
feat(periodic_scheduler): support generic git ref

hamir-suspect:
## 📝 Description
  - Adds support for scheduling periodic workflows on Git tags in addition to branches
  - Introduces a new git_reference utility module to handle both branch and tag references
  - Updates the periodic scheduler to accept and process tag-based schedules
  - **gRPC/Protocol buffer updates:** Regenerated protocol buffer definitions to rename `branch` field into `reference`, will still work if caller service uses `branch` field name ([since type and order in api is preserved](https://stackoverflow.com/questions/45431685/protocol-buffer-does-changing-field-name-break-the-message))

## ✅ Checklist

### DIFF
--- a/periodic_scheduler/scheduler/lib/scheduler/actions/schedule_wf_impl.ex
+++ b/periodic_scheduler/scheduler/lib/scheduler/actions/schedule_wf_impl.ex
@@ -154,7 +154,7 @@ defmodule Scheduler.Actions.ScheduleWfImpl do
     |> legacy_branch_name_format()
   end
 
-  defp legacy_branch_name_format("refs/tags/" <> _ = full_ref), do: full_ref
+  defp legacy_branch_name_format(full_ref = "refs/tags/" <> _), do: full_ref
 
   defp legacy_branch_name_format("refs/pull/" <> rest) do
     pr_number = rest |> String.trim_trailing("/head")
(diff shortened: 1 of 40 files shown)


periodic_scheduler/scheduler/lib/scheduler/actions/schedule_wf_impl.ex
```diff
@@ -117,18 +86,24 @@ defmodule Scheduler.Actions.ScheduleWfImpl do
...
+    # Handle backwards compatibility: normalize reference to full format
+    git_reference = GitReference.normalize(trigger.reference)
+
+    # Extract branch name for legacy branch_name field
+    branch_name = GitReference.extract_name(trigger.reference)
+
     %{
       service: schedule_workflow_service_type(repository.integration_type),
-      repo: %{branch_name: trigger.branch},
+      repo: %{branch_name: branch_name},
```
DamjanBecirovic, line 100: Plumber expects branch_name here to be in the following format:
- for branches: just the normalized branch name as it is here currently
- for tags: full reference, e.g. `refs/tags/v1.2.3`
- for PRs, specific format: `pull-request-<pull request number>`

This is legacy behavior initially set by repo-proxy/github_hooks service, and the reason why we later  introduced the `label` field, which has normalized names for all three types.

hamir-suspect: `branch_name` now should comply with the format

skipi (approved):

DamjanBecirovic (approved):

PR #584, opened 2025-09-17
feat(public-api): Support tags for tasks/periodics

hamir-suspect:
## 📝 Description
  - Added tag support to task/periodic scheduler endpoints: Create, update, replace, and trigger operations now accept an optional tags parameter
  - Extended internal gRPC clients: Updated request/response formatters to handle tags when communicating with the periodic scheduler service
  - Updated OpenAPI schemas: Added tag field definitions to task specifications and trigger specifications
  
## ✅ Checklist

### DIFF
--- a/public-api/v1alpha/lib/pipelines_api/periodic_scheduler_client/response_formatter.ex
+++ b/public-api/v1alpha/lib/pipelines_api/periodic_scheduler_client/response_formatter.ex
@@ -167,15 +167,8 @@ defmodule PipelinesAPI.PeriodicSchedulerClient.ResponseFormatter do
     end
   end
 
-  def reference_to_branch_field(reference) when is_binary(reference) do
-    with true <- String.starts_with?(reference, "refs/heads/"),
-         true <- not String.starts_with?(reference, "refs/pull/"),
-         true <- not String.starts_with?(reference, "refs/tags/") do
-      String.replace_prefix(reference, "refs/heads/", "")
-    else
-      _ -> nil
-    end
-  end
+  def reference_to_branch_field("refs/heads/" <> branch_name), do: branch_name
+  def reference_to_branch_field(_), do: nil
 
   defp log_invalid_response(response, rpc_method) do
     response


public-api/v1alpha/lib/pipelines_api/periodic_scheduler_client/request_fromatter.ex
```diff
@@ -121,6 +124,24 @@ defmodule PipelinesAPI.PeriodicSchedulerClient.RequestFormatter do
...
 
+  defp build_reference(params) do
+    case params |> Map.get("reference") do
+      reference_map when is_map(reference_map) ->
+        reference_type = Map.get(reference_map, "type", "BRANCH")
+        reference_name = Map.get(reference_map, "name", "")
+
+        case String.upcase(reference_type) do
+          "TAG" -> "refs/tags/#{reference_name}"
+          _ -> "refs/heads/#{reference_name}"
```
DamjanBecirovic, line 135: This is now missing the `PR` case.

public-api/v1alpha/lib/pipelines_api/periodic_scheduler_client/response_formatter.ex
```diff
@@ -152,6 +156,27 @@ defmodule PipelinesAPI.PeriodicSchedulerClient.ResponseFormatter do
...
+    case reference_to_branch_field(periodic.reference) do
+      nil -> periodic
+      branch_name -> Map.put(periodic, :branch, branch_name)
+    end
+  end
+
+  def reference_to_branch_field(reference) when is_binary(reference) do
+    with true <- String.starts_with?(reference, "refs/heads/"),
+         true <- not String.starts_with?(reference, "refs/pull/"),
+         true <- not String.starts_with?(reference, "refs/tags/") do
```
DamjanBecirovic: The last two `String.starts_with?` checks are redundant, since if the first one is true, the rest will always be false.

public-api/v1alpha/lib/pipelines_api/periodic_scheduler_client/request_fromatter.ex
hamir-suspect: I didn't add `PR` in the UI so I removed it here as well, we can easily add it if we get a FR. This also prevented PR case https://github.com/semaphoreio/semaphore/pull/584/files#diff-08b07813ddaf1b74e7b39d039b97e08c5190b4a6deaeece94c5653d2c786a8ecR108

Issue #630, opened 2025-10-07
[DOCS] Scattered info on running jobs in containers


markoa:
Semaphore can run jobs in custom Docker containers, which is a way for organizations to bring their own pre-built environment. However it is close to impossible to discover this by browsing documentation.

Information is scattered in the following few places and does not show up easily in search:
- How to use it https://docs.semaphore.io/using-semaphore/pipelines#docker-environments
- How to use an image from a private registry https://docs.semaphore.io/using-semaphore/pipelines#docker-private
- Reference in yaml https://docs.semaphore.io/reference/pipeline-yaml#containers

Proposed solution is to start by re-composing the content to appear in the top-level _Containers_ section.

PR #634, opened 2025-10-08
docs: move docker and container documentation to top-level page

TomFern:
## 📝 Description
Move the documentation of Docker and Containers to the top level page in the sidebar. Reorganizes information about Docker and Container in one place so it's easy to find and search.

Addresses https://github.com/semaphoreio/semaphore/issues/630

## ✅ Checklist

### DIFF
--- a/docs/docs/using-semaphore/containers.md
+++ b/docs/docs/using-semaphore/containers.md
@@ -2,7 +2,7 @@
 description: Using Docker and Containers in Semaphore
 ---
 
-# Use Docker for Running Jobs
+# Running Jobs in Docker Containers
 
 Run jobs inside Docker containers orchestrated by Semaphore.
 
--- a/docs/docs/using-semaphore/containers/docker.md
+++ b/docs/docs/using-semaphore/containers/docker.md
@@ -9,7 +9,7 @@ Use Semaphore to build, test, store, and deploy Docker images to production. Thi
 
 :::tip
 
-This page describes how to build, test, and publish Docker images using Semaphore. If you want to run [jobs](../jobs) inside Docker containers, see the [Docker and Containers page](../containers).
+This page describes how to build, test, and publish Docker images using Semaphore. If you want to run [jobs](../jobs) inside Docker containers, see the [Running Jobs in Docker Containers page](../containers).
 
 :::
 
--- a/docs/docs/using-semaphore/pipelines.md
+++ b/docs/docs/using-semaphore/pipelines.md
@@ -237,7 +237,7 @@ blocks:
 
 ### Docker containers {#docker-environments}
 
-The [Docker and Containers page](./containers) shows how to use Docker environments in your jobs.
+The [Running Jobs in Docker Containers page](./containers) shows how to use Docker environments in your jobs.
 
 
 ### Prologue {#prologue}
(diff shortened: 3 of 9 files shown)


docs/docs/using-semaphore/containers.md
```diff
@@ -0,0 +1,255 @@
...
+mysql --host=mysql --user=root
+```
+
+## How to run jobs inside Docker containers
+
+To run the job inside a Docker container you must define at a Docker image in your agent configuration.
+
+The following example shows how to set up two containers:
+
+- main: where the commands of the job are executed. The container is running an Ubuntu 20.22 container. The image is pulled from the [Semaphore Container Registry](./containers/container-registry)
```
DamjanBecirovic: Ubuntu 20.22 -> Ubuntu 20.04

docs/docs/using-semaphore/containers.md
```diff
@@ -0,0 +1,255 @@
+---
+description: Using Docker and Containers in Semaphore
+---
+
+# Docker and Containers
```
DamjanBecirovic: I would suggest changing this to something like `Run jobs in containers`, or `Use Docker for running jobs`, or even `Jobs in containers` if it needs to be extremely brief.
Connection to jobs is not really clear from the current version IMO, especially if examining it without full page context, like in the navigation bar or search results.

docs/docs/using-semaphore/containers.md
TomFern: good catch

docs/docs/using-semaphore/containers.md
TomFern: Done. Thank you @DamjanBecirovic

docs/sidebars.js
```diff
@@ -132,10 +132,13 @@ const sidebars = {
           },
         ],
       },
-
       {
         type: 'category',
-        label: 'Containers',
+        label: 'Docker and Containers',
```
DamjanBecirovic, line 137: This should be updated to use the new title of the page.

docs/docs/using-semaphore/pipelines.md
```diff
@@ -237,91 +237,8 @@ blocks:
 
 ### Docker containers {#docker-environments}
 
-:::tip
+The [Docker and Containers page](./containers) shows how to use Docker environments in your jobs.
```
DamjanBecirovic: This should be updated to use the new title of the page. The same goes for the other versioned copies of this file.

docs/docs/using-semaphore/containers/docker.md
```diff
@@ -5,17 +5,11 @@ sidebar_position: 5
...
-
-
-
-
 Use Semaphore to build, test, store, and deploy Docker images to production. This page explains how to use Docker inside Semaphore.
 
 :::tip
 
-This page describes how to build, test, and publish Docker images using Semaphore. If you want to run [jobs](../jobs) inside Docker containers, see the [Docker environments page](../pipelines#docker-environments).
+This page describes how to build, test, and publish Docker images using Semaphore. If you want to run [jobs](../jobs) inside Docker containers, see the [Docker and Containers page](../containers).
```
DamjanBecirovic: This should be updated to use the new title of the page. The same goes for the other versioned copies of this file.

docs/sidebars.js
TomFern: This is not the title of the page but the title of the section. The section has 3 pages in total: Run Jobs in Docker Container, Building Docker Images and the Semaphore Docker Registry

DamjanBecirovic: Ah, I see, I saw the same title and assumed that it was for a subsection. Thanks for the info 👍

PR #645, opened 2025-10-10
fix(front): test suite duration/sorting

marco2216:
## 📝 Description
Test suite durations are not necessarily the sum of the tests that make up the suite. If using Jest, in most cases the test suite time is longer than the sum of the individual tests.
There's two changes here because
1. I would argue that the test suite duration should not be changed even when hiding some tests.
2. suite.syncSummary was being called which mutates the suite object, so the suite duration was being updated after the sorting happened, which means that the sorting was inconsistent with the rendered durations in the UI.

You can see the issue here - sorting is wrong. I found the issue by comparing with the junit.xml test results file.
image

Also, as a general comment, it's a very bad idea to mutate things when writing React code, it makes it extremely hard to understand what happens with state. https://react.dev/learn/updating-objects-in-state#treat-state-as-read-only :)

## ✅ Checklist
- [ ] I have tested this change (can't figure out how to run this locally, but I'm pretty sure this is a fix - would be great if you could test it)


hamir-suspect (approved):
Hey @marco2216 looking back at this, I'm actually ok with merging as is.

skipi (approved):

PR #664, opened 2025-10-22
feat(front): add members download button

loadez:
## 📝 Description
- Add members download button on people page
- Implement export members as csv endpoint
- Add export members test

## ✅ Checklist

### DIFF
--- a/front/test/front_web/controllers/service_account_controller_test.exs
+++ b/front/test/front_web/controllers/service_account_controller_test.exs
@@ -268,6 +268,65 @@ defmodule FrontWeb.ServiceAccountControllerTest do
     end
   end
 
+  describe "GET /service_accounts/export" do
+    test "requires service_accounts.view permission", %{
+      conn: conn,
+      org_id: org_id,
+      user_id: user_id
+    } do
+      Support.Stubs.PermissionPatrol.remove_all_permissions()
+      Support.Stubs.PermissionPatrol.add_permissions(org_id, user_id, ["organization.view"])
+
+      conn = get(conn, "/service_accounts/export")
+
+      assert html_response(conn, 404) =~ "Page not found"
+    end
+
+    test "when the user can access the org => send csv", %{
+      conn: conn,
+      org_id: org_id,
+      user_id: user_id
+    } do
+      Support.Stubs.PermissionPatrol.add_permissions(org_id, user_id, ["organization.view"])
+
+      expect(ServiceAccountMock, :describe_many, fn members ->
+        service_accounts =
+          Enum.map(members, fn member_id ->
+            %InternalApi.ServiceAccount.ServiceAccount{
+              id: member_id,
+              name: "Test Service Account",
+              description: "Test description",
+              org_id: org_id,
+              creator_id: "",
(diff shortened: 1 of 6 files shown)


front/test/front_web/controllers/people_controller_test.exs
```diff
@@ -36,6 +36,20 @@ defmodule FrontWeb.PeopleControllerTest do
...
+      PermissionPatrol.remove_all_permissions()
+
+      conn =
+        conn
+        |> get("/people/export")
+
+      assert html_response(conn, 404) =~ "404"
+    end
+  end
+
```
skipi, line 78: Let's add one more successful test case here. We should check the fields of the CSV export

front/test/front_web/controllers/service_account_controller_test.exs
```diff
@@ -268,6 +268,21 @@ defmodule FrontWeb.ServiceAccountControllerTest do
...
+      user_id: user_id
+    } do
+      Support.Stubs.PermissionPatrol.remove_all_permissions()
+      Support.Stubs.PermissionPatrol.add_permissions(org_id, user_id, ["organization.view"])
+
+      conn = get(conn, "/service_accounts/export")
+
+      assert html_response(conn, 404) =~ "Page not found"
+    end
+  end
```
skipi, line 328: Let's add one more successful test case here. We should check the fields of the CSV export

front/lib/front_web/controllers/people_controller.ex
```diff
@@ -1061,6 +1065,51 @@ defmodule FrontWeb.PeopleController do
...
+      page_size = 100
+
+      data =
+        Stream.unfold(0, fn
+          nil ->
+            nil
+
+          page_no ->
+            case Members.list_org_members(org_id, page_no: page_no, page_size: page_size) do
+              {:ok, {[_ | _] = members, total_pages}} ->
```
skipi: I think this match `[_ | _] =` is not needed here. It ensures that there is at least one element in the members list, but that's not important here.

front/test/front_web/controllers/people_controller_test.exs
loadez: Done, now we are checking the csv header fields and row length.

front/lib/front_web/controllers/people_controller.ex
loadez: Nice catch, removed.

The next page function will return nil when we exceed the total pages value so the stream unfold is ended in the next interation.

front/test/front_web/controllers/people_controller_test.exs
```diff
@@ -36,6 +36,46 @@ defmodule FrontWeb.PeopleControllerTest do
...
+      assert response_content_type(conn, :csv)
+
+      rows =
+        conn.resp_body
+        |> String.split("\r\n", trim: true)
+        |> CSV.decode!(validate_row_length: true, headers: true)
+        |> Enum.to_list()
+
+      assert length(rows) > 0
```
skipi: Let's assert the expected number of rows here instead of just > 0 - it will always be true(the list will have at least 1 row, because the header is counted as well).

skipi (approved):

PR #677, opened 2025-10-27
feat: github app can be created on organization and can be made public

hamir-suspect:
## 📝 Description
Connect button for github app instead of going straight to github with our github app manifest will prompt for a few possible modifications of manifest. This enables users using CE/EE installations to easily create github app on their organization instead of (previously hardcoded) personal account. Also allows making the github app public.

image

## ✅ Checklist

### DIFF
--- a/guard/lib/guard/instance_config/api.ex
+++ b/guard/lib/guard/instance_config/api.ex
@@ -227,7 +227,7 @@ defmodule Guard.InstanceConfig.Api do
 
   defp github_app_install_url(org) when is_binary(org) and byte_size(org) > 0 do
     base_url = Application.get_env(:guard, :github_app_base_url, "https://github.com")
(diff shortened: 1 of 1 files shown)


front/assets/js/git_integration/pages/github_app_setup.tsx
```diff
@@ -0,0 +1,150 @@
...
+                    id="organizationName"
+                    type="text"
+                    value={organizationName}
+                    onChange={(e) => setOrganizationName((e.target as HTMLInputElement).value)}
+                    placeholder="e.g., my-company"
+                    className="form-control w-100"
+                    required
+                  />
+                  <p className="f6 gray mt2 mb0">
+                    The exact name of your GitHub organization
```
radwo, line 109: Please extend this line with info that you need to be an admin on this organization.

PR #693, opened 2025-11-04
MCP: Add feature flag checks

DamjanBecirovic:
## 📝 Description

Changes:
- Adds infrastructure for checking feature flags based on the one from the `velocity` service
- Reorganizes mocks and stubs to avoid cycle import issues
- Adds a check for `mcp_server_read_tools` feature flag before any tool execution

## ✅ Checklist
- ~[x] This change requires a documentation update~

### DIFF
--- a/mcp_server/pkg/internalapi/config.go
+++ b/mcp_server/pkg/internalapi/config.go
@@ -134,9 +134,6 @@ func (c Config) Validate() error {
 	if c.RBACEndpoint == "" {
 		missing = append(missing, "rbac gRPC endpoint")
 	}
-	if c.RBACEndpoint == "" {
-		missing = append(missing, "rbac gRPC endpoint")
-	}
 	if c.FeatureHubEndpoint == "" {
 		missing = append(missing, "feature hub gRPC endpoint")
 	}


mcp_server/pkg/internalapi/config.go
```diff
@@ -67,6 +71,7 @@ type Config struct {
 	Loghub2Endpoint      string
 	UserEndpoint         string
 	RBACEndpoint         string
+	FeatureHubEndpoint   string
```
dexyk, line 74: @DamjanBecirovic don't see this config value being updated in `LoadConfig()` below

mcp_server/pkg/internalapi/config.go
```diff
@@ -54,6 +54,10 @@ var (
 		"INTERNAL_API_URL_RBAC",
 		"MCP_RBAC_GRPC_ENDPOINT",
 	}
+	FeatureHubEndpoint = []string{
```
dexyk: @DamjanBecirovic this is not used in `LoadConfig()` below and we should probably follow the above pattern to name this const  like `rbacEndpointEnvs` etc

dexyk (changes requested):

mcp_server/pkg/internalapi/config.go
DamjanBecirovic: The name pattern is fixed.

mcp_server/pkg/internalapi/config.go
DamjanBecirovic: I've fixed this issue, added tests, and verified that it works on the preprod environment.

PR #708, opened 2025-11-11
fix(periodic_scheduler): handle partial unsuspend failures

dexyk:
## 📝 Description

Keep unsuspending batches even when individual periodics fail, and surface failures via logging.

Related [task](https://github.com/renderedtext/tasks/issues/8906).

## ✅ Checklist

### DIFF
--- a/periodic_scheduler/scheduler/lib/scheduler/events_consumers/org_unblocked.ex
+++ b/periodic_scheduler/scheduler/lib/scheduler/events_consumers/org_unblocked.ex
@@ -36,35 +36,60 @@ defmodule Scheduler.EventsConsumers.OrgUnblocked do
   def unsuspend_periodics_from_org(error),
     do: error |> LT.warn("Error while processing org unblocked RabbitMQ message:")
 
-  defp unsuspend_batch(org_id, batch_no) do
+  defp unsuspend_batch(org_id, batch_no, has_failures \\ false) do
     with {:ok, periodics} <- PeriodicsQueries.get_all_from_org(org_id, batch_no),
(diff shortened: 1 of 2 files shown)


hamir-suspect:
@codex review

periodic_scheduler/scheduler/lib/scheduler/events_consumers/org_unblocked.ex
```diff
@@ -51,13 +59,27 @@ defmodule Scheduler.EventsConsumers.OrgUnblocked do
...
+          {:ok, periodic} ->
+            %{acc | unsuspended: [periodic | acc.unsuspended]}
+
+          {:error, reason} ->
+            failed_entry = %{id: periodic.id, reason: reason}
+            %{acc | failed: [failed_entry | acc.failed]}
+
+          other ->
+            failed_entry = %{id: periodic.id, reason: other}
+            %{acc | failed: [failed_entry | acc.failed]}
```
DamjanBecirovic, line 96: With previous logic, the worker would fail if there were any issues with unsuspending a periodic, and then the supervisor would restart it to try again, following the "let it crash" Elixir/Erlang approach. That allowed the system to recover from any kind of transient issue in the unsuspending periodics logic, like DB being temporarily unavailable, network issues, etc.

With this new approach, we are no longer resistant to those temporary issues since all errors are now just logged and metrics are sent, but there is no retry logic.

Rather than catching all errors here, it should follow the new logic of ignoring + logging and metrics only in cases when we know exactly that the issue is not transient, i.e., when the error is `{:error, :missing_cron_expression}`, and by default, it should follow the old logic of crashing and retrying.

PR #713, opened 2025-11-13
feat(zebra): create job deletion policy worker

loadez:
## 📝 Description

This pr is mainly for the marker worker. If the deletion worker needs any modifications (to tell artifacthub and loghub that the job was deleted) it will be done in another pr to keep this one simple.

Currently the deletion worker is capable of correctly deleting data from jobs.

## ✅ Checklist


loadez:
@codex review

TomFern (dismissed):

github_hooks/db/migrate/20251119104017_add_organization_created_index_at_jobs_table.rb
```diff
@@ -0,0 +1,11 @@
+class AddOrganizationCreatedIndexAtJobsTable < ActiveRecord::Migration[6.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :jobs,
```
dexyk, line 5: @loadez please check could the extended time required for the CONCURRENTLY index build lead to liveness/readiness Probe failure (as Puma is not running), resulting in Kubernetes terminating the pod before completion? Should we run this as a dedicated Job?

loadez: This is a valid concern I will raise this question to the rest of the team. In the other migrations that add indexes we don't do anything special, but for jobs we risk a timeout depending on the configuration of the database.

loadez: I validated with the team and we will run this migration with extended timeouts.

loadez:
@codex review

zebra/lib/zebra/models/job.ex
```diff
@@ -356,6 +357,68 @@ defmodule Zebra.Models.Job do
...
 
+  def mark_jobs_for_deletion(org_id, cutoff_date, deletion_days) do
+    import Ecto.Query, only: [from: 2]
+
+    query =
+      from(j in Zebra.Models.Job,
+        where:
+          is_nil(j.expires_at) and
+            j.organization_id == ^org_id and
+            j.created_at <= ^cutoff_date,
```
skipi, line 368: This piece is missing: https://github.com/semaphoreio/semaphore/pull/717/files#diff-8a63d967374ed19cbae376be1fa028d406a1c2d87ac399a2d5e70adc620b07f8R38-R47

github_hooks/db/migrate/20251118091829_add_expires_at_to_jobs.rb
```diff
@@ -0,0 +1,5 @@
+class AddExpiresAtToJobs < ActiveRecord::Migration[6.1]
+  def change
+    add_column :jobs, :expires_at, :datetime
```
skipi: We need to add if not exists here as well.

PR #717, opened 2025-11-18
feat(plumber): retention policy worker

skipi:
## 📝 Description

This PR implements retention policy support in plumber. When the usage service applies a retention policy, plumber marks pipeline records for expiration with a configurable grace period, allowing corrections before data is deleted.

- Added `expires_at` column to `pipeline_requests` table to track when records should be deleted
- Created `PolicyConsumer` to receive retention policy events from usage service via RabbitMQ
- Created `PolicyApplier` to mark/unmark pipeline records based on the cutoff date
- Records older than cutoff are marked with `expires_at = now + grace_period` (default 15 days)
- Records newer than the cutoff have `expires_at` cleared, allowing policy corrections to "save" mistakenly marked records
- Grace period is configurable via `RETENTION_GRACE_PERIOD_DAYS` env var (min: 7 days)
- Record Deleter Worker - Added a GenServer that periodically deletes expired records in batches

## ✅ Checklist

### DIFF
--- a/plumber/ppl/config/config.exs
+++ b/plumber/ppl/config/config.exs
@@ -43,7 +43,6 @@ config :gofer_client, gofer_grpc_timeout: 4_567
 # Time in hours before delete requests are processed
 config :ppl, deletion_offset_h: 24
 
-
 # How many times should wormhole retry to publish pipeline events to RabbitMQ
 config :ppl, publish_retry_count: 3
 # Timeout for publishing pipeline events to RabbitMQ
--- a/plumber/ppl/config/runtime.exs
+++ b/plumber/ppl/config/runtime.exs
@@ -41,9 +41,7 @@ config :watchman,
 
 # Retention policy event consumer
 config :ppl, Ppl.Retention.PolicyConsumer,
-  enabled: System.get_env("RETENTION_CONSUMER_ENABLED", "false") == "true",
-  exchange: System.get_env("USAGE_POLICY_EXCHANGE", "usage_internal_api"),
-  routing_key: System.get_env("USAGE_POLICY_ROUTING_KEY", "usage.apply_organization_policy")
+  enabled: System.get_env("RETENTION_CONSUMER_ENABLED", "false") == "true"
 
 # Retention policy applier settings
 config :ppl, Ppl.Retention.PolicyApplier,
--- a/plumber/ppl/config/test.exs
+++ b/plumber/ppl/config/test.exs
@@ -47,10 +47,6 @@ config :ppl, dr_pending_sp: 100
 config :ppl, dr_deleting_sp: 100
 config :ppl, dr_queue_deleting_sp: 100
 
-config :ppl, Ppl.Retention.PolicyConsumer,
-  exchange: "usage_internal_api_test",
-  routing_key: "usage.apply_organization_policy.test"
-
 config :watchman,
   host: "localhost",
   port: 8125,
--- a/plumber/ppl/lib/ppl/retention/policy_consumer.ex
+++ b/plumber/ppl/lib/ppl/retention/policy_consumer.ex
@@ -3,12 +3,10 @@ defmodule Ppl.Retention.PolicyConsumer do
   Subscribes to usage.ApplyOrganizationPolicyEvent and marks pipelines for expiration.
   """
 
-  @consumer_opts Application.compile_env(:ppl, __MODULE__, [])
-
   use Tackle.Consumer,
     url: System.get_env("RABBITMQ_URL"),
-    exchange: Keyword.get(@consumer_opts, :exchange, "usage_internal_api"),
-    routing_key: Keyword.get(@consumer_opts, :routing_key, "usage.apply_organization_policy"),
+    exchange: "usage_internal_api",
+    routing_key: "usage.apply_organization_policy",
     service: "plumber-retention"
 
   require Logger


skipi:
@codex review

loadez (approved):

plumber/ppl/lib/ppl/retention/record_deleter_queries.ex
```diff
@@ -0,0 +1,50 @@
...
+  def delete_expired_batch(limit) do
+    now = NaiveDateTime.utc_now()
+
+    expired_records =
+      from(pr in PplRequests,
+        where: pr.expires_at < ^now,
+        select: %{id: pr.id, ppl_artefact_id: pr.ppl_artefact_id},
+        limit: ^limit
+      )
+      |> EctoRepo.all()
```
loadez: Just a comment, but try to run this with the explain to validate that the planner picks the index that you created to run this query. (It should infer expires_at is not null but it may not)

skipi: I've added additional index for this:

https://github.com/semaphoreio/semaphore/pull/717/files#diff-43f618970cd9f390142fc6904395205cb769a1600d895c07f5132c043c34516eR14

plumber/ppl/lib/ppl/retention/policy_applier.ex
```diff
@@ -0,0 +1,65 @@
...
+    {marked_count, unmarked_count}
+  end
+
+  defp grace_period_days do
+    config = Application.get_env(:ppl, __MODULE__, [])
+    days = Keyword.get(config, :grace_period_days, @default_grace_period_days)
+
+    if days < @min_grace_period_days do
+      Logger.warning(
+        "[Retention] grace_period_days=#{days} is below minimum, using #{@min_grace_period_days}"
```
DamjanBecirovic: Let's add `org_id` in this log line, so it is easier to debug.

plumber/ppl/lib/ppl/retention/policy_applier.ex
```diff
@@ -0,0 +1,65 @@
...
+
+    unmark_query =
+      from(pr in PplRequests,
+        where: fragment("?->>?", pr.request_args, "organization_id") == ^org_id,
+        where: pr.inserted_at >= ^cutoff,
+        where: not is_nil(pr.expires_at)
+      )
+
+    {marked_count, _} = EctoRepo.update_all(mark_query, set: [expires_at: expires_at])
+    {unmarked_count, _} = EctoRepo.update_all(unmark_query, set: [expires_at: nil])
```
DamjanBecirovic: These `update_all` calls can cause issues since if the cutoff date is recent enough, we might end up updating a lot of rows at once and have issues with locks, heavy IO, connection timeout, etc.

The more secure way here would be to do batching with a limit on the number of rows that are updated at once.

plumber/ppl/lib/ppl/retention/policy_applier.ex
```diff
@@ -0,0 +1,110 @@
...
+    ids_query =
+      from(pr in PplRequests,
+        where: fragment("?->>?", pr.request_args, "organization_id") == ^org_id,
+        where: pr.inserted_at < ^cutoff,
+        where: is_nil(pr.expires_at),
+        select: pr.id,
+        limit: ^batch_size
+      )
+
+    ids = EctoRepo.all(ids_query)
```
radwo: Why do you have this extra query here? can't you just call `update_all` with `pr.id in subquery`? if it work it will be one less trip to DB

skipi: No reason - fixed.

plumber/ppl/lib/ppl/retention/record_deleter_queries.ex
```diff
@@ -0,0 +1,50 @@
...
+
+    case expired_records do
+      [] ->
+        {:ok, 0}
+
+      records ->
+        Enum.each(records, fn %{ppl_artefact_id: ppl_id} ->
+          Block.delete_blocks_from_ppl(ppl_id)
```
DamjanBecirovic: We should not use here `ppl_artefact_id` since in the case of a partially rebuilt pipeline, it will have the value of the original pipeline's ppl_id, so the blocks from the rebuilt pipeline will be left in the database.

Instead, you should use the `id` field from the select query.

plumber/ppl/test/retention/record_deleter_queries_test.exs
```diff
@@ -0,0 +1,125 @@
...
+  alias Ppl.EctoRepo
+  alias Ppl.PplRequests.Model.PplRequests
+  alias Ppl.Retention.RecordDeleterQueries
+
+  setup do
+    Test.Helpers.truncate_db()
+    :ok
+  end
+
+  describe "delete_expired_batch/1" do
```
DamjanBecirovic, line 19: Can you add one integration test here that does the following:
- starts a pipeline execution and waits for it to finish successfully, [example](https://github.com/semaphoreio/semaphore/blob/main/plumber/ppl/test/grpc/server_test.exs#L511-L518) 
- deletes the pipeline using this new query
- asserts that records are deleted from all related pipeline tables similar to [this](https://github.com/semaphoreio/semaphore/blob/main/plumber/ppl/test/ppl_requests/model/ppl_requests_queries_test.exs#L136), and that related blocks are deleted in the block_repo database

DamjanBecirovic (approved):

DamjanBecirovic (approved):

PR #733, opened 2025-11-26
feat(mcp_server): read test result junit

hamir-suspect:
## 📝 Description
- Add MCP `get_test_results` tool to return signed URLs for pipeline/job JUnit artifacts with org/project/workflow validation and authz checks
- Wire artifacthub proto/module and register tool in server bootstrap; extend internal API tooling to request signed URLs for junit path
- Add mocks and unit tests covering public/private project access, RBAC, and ID validation for both scopes
- **Add MCP prompts capability** with `semaphore_agent_setup` prompt that provides AI agents with optimal usage guidance:
  - Organization/project ID discovery and caching workflow
  - Test results best practice: download signed URL to local file once, analyze locally (URLs expire quickly)
  - Debugging workflow patterns (workflows → pipelines → jobs → logs → test results)
  - Complete tools reference with key parameters
  - Caching strategy to minimize API calls
- **Add fallback artifact resolution**: tries `mcp-summary.json` first, falls back to `junit.xml` (jobs) or gzipped summary (pipelines) if primary format unavailable

## ✅ Checklist

### DIFF
--- a/periodic_scheduler/scheduler/AGENTS.md
+++ b/periodic_scheduler/scheduler/AGENTS.md
@@ -1,0 +1,23 @@
+# Repository Guidelines
+
+## Project Structure & Module Organization
+Runtime code lives under `lib/`, split by domain boundaries (`scheduler/`, `actions/`, `workers/`). Persistence details sit in `lib/scheduler/periodics` and `lib/scheduler/front_db` with schemas mirroring tables. Configuration defaults are in `config/config.exs`, with environment overrides in `config/{dev,test,prod}.exs`. Database migrations and seeds are under `priv/{periodics_repo,front_repo}/`, while `test/` mirrors `lib/` one-to-one for ExUnit coverage. Docker, release, and deployment assets reside in `docker-compose.yml`, `rel/`, and `helm/`. Shared automation scripts sit inside `scripts/`. For a deeper walkthrough, treat `DOCUMENTATION.md` as the go-to triage guide before diving into tasks.
+
+## Build, Test, and Development Commands
+- `mix deps.get && mix compile` installs dependencies and compiles the application locally.
+- `MIX_ENV=test make test.ex.setup` prepares the Postgres schema inside Docker for integration tests.
+- `mix test` or `mix test --cover` runs the ExUnit suite (JUnit formatter available via `MIX_ENV=test mix test --formatter JUnitFormatter` for CI).
+- `mix credo --strict` runs linting, and `mix format` enforces the formatter prior to commits.
+- `docker compose up app` boots the scheduler in a container with RabbitMQ/Postgres defaults from the `.env` files.
+
+## Coding Style & Naming Conventions
+Follow the default Elixir formatter (2-space indentation, pipe-first style). Modules use `Scheduler.*` namespaces that map to folders (e.g., `Scheduler.Periodic.Job` ⇔ `lib/scheduler/periodic/job.ex`). Functions are snake_case verbs, macros are camel-case nouns, and tests end in `_test.exs`. Keep public modules documented with `@moduledoc` and prefer pattern matching + guard clauses over nested conditionals. Run `mix format && mix credo` before every push.
+
+## Testing Guidelines
+Author unit tests alongside code in `test/<mirror_path>_test.exs`. Use ExUnit’s `describe` blocks per function and tag integration tests with `@moduletag :integration` so CI can filter. Ensure new DB queries include data factories from `test/support`. Aim to maintain or improve coverage reported by `mix test --cover`; add regression tests for every bugfix.
+
+## Commit & Pull Request Guidelines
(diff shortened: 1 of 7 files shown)


mcp_server/pkg/prompts/prompts.go
```diff
@@ -0,0 +1,357 @@
...
+	)
+}
+
+func agentSetupHandler() server.PromptHandlerFunc {
+	return func(ctx context.Context, request mcp.GetPromptRequest) (*mcp.GetPromptResult, error) {
+		projectName := ""
+		if args := request.Params.Arguments; args != nil {
+			if name, exists := args["project_name"]; exists {
+				projectName = name
+			}
```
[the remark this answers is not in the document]
hamir-suspect: This comment is incorrect. `GetPromptParams.Arguments` is typed as `map[string]string`, not `map[string]any`. See: https://pkg.go.dev/github.com/mark3labs/mcp-go/mcp#GetPromptParams

The code compiles and all tests pass:
```
go build ./...  # ✓
go test ./pkg/prompts/... # ✓
```

mcp_server/pkg/tools/testresults/testresults.go
```diff
@@ -0,0 +1,571 @@
...
+			mcp.Pattern(`^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`),
+		),
+		mcp.WithString(
+			"pipeline_id",
+			mcp.Description("Pipeline UUID (required when scope=pipeline). Get this from pipelines_list. Format: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx."),
+			mcp.Pattern(`^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`),
+		),
+		mcp.WithString(
+			"workflow_id",
+			mcp.Description("Workflow UUID that owns the pipeline (required when scope=pipeline). Get this from workflows_search. Format: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx."),
```
DamjanBecirovic: This should not be required or a parameter at all, since we can infer it from pipeline_id by calling pipeline describe.

mcp_server/pkg/prompts/prompts.go
```diff
@@ -0,0 +1,146 @@
...
+
+import (
+	"context"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+const (
+	agentSetupPromptName = "semaphore_agent_setup"
```
DamjanBecirovic: What do you think about calling this `semaphore_mcp_setup` or `semaphore_mcp_init`?

mcp_server/pkg/prompts/prompts.go
```diff
@@ -0,0 +1,146 @@
...
+` + "```" + `
+
+Discover IDs once with ` + "`organizations_list`" + ` and ` + "`projects_list`" + `, then always use cached values.
+
+## Debugging Workflow
+
+1. ` + "`workflows_search`" + ` → find failing workflow
+2. ` + "`pipelines_list`" + ` → get pipeline from workflow
+3. ` + "`pipeline_jobs`" + ` → find failed jobs
+4. ` + "`jobs_logs`" + ` → read error output
```
DamjanBecirovic: Here we could instruct it to:
- if pipeline failed with `result_reason=test`, try to fetch test results first, and only check logs if there are no test results 
- otherwise, fetch job logs

mcp_server/pkg/tools/testresults/testresults.go
```diff
@@ -0,0 +1,540 @@
...
+		if message == "" {
+			message = "pipeline describe returned non-OK status"
+		}
+		return nil, fmt.Errorf("pipeline describe failed: %s", message)
+	}
+	if resp.GetPipeline() == nil {
+		return nil, fmt.Errorf("pipeline describe returned no pipeline payload")
+	}
+	return resp.GetPipeline(), nil
+}
```
DamjanBecirovic: We have this same logic for `describePipeline(...)` in [pipeline_jobs](https://github.com/semaphoreio/semaphore/blob/main/mcp_server/pkg/tools/pipelines/pipelines.go#L498-L526) tool implementation, and `describeProject(...)` is already implemented in a similar way in [workflow_run](https://github.com/semaphoreio/semaphore/blob/main/mcp_server/pkg/tools/workflows/run_tool.go#L162-L182) tool implementation.

As we add more tools, we will probably need those again sooner or later. 

The best solution here is to add a new package called `clients` or something similar and implement these API calls there, and then refactor the tools to use that shared package.

This can also be done in a separate PR if that works better.

mcp_server/pkg/prompts/prompts.go
hamir-suspect: when calling the prompt from the MCP it's usually prefixed by MCP name, so we can just call it `mcp_setup` ?

PR #764, opened 2025-12-12
fix(front): allow assigning project roles to service accounts

hamir-suspect:
## Summary

- Allows service accounts to be added to projects even when `rbac__project_roles` feature flag is disabled
- Adds new `show_service_account_management_buttons?/3` function that bypasses the feature flag check for service accounts at project level
- Regular users still require the feature flag for project-level role management

Closes https://github.com/renderedtext/tasks/issues/8995

## Test plan

- [x] Without `rbac__project_roles` flag enabled:
  - Navigate to project people page
  - Verify "Add service accounts" button shows and popup opens
  - Add a service account to project
  - Verify "Change role" and "Remove" buttons appear for service accounts
  - Verify regular users don't have management buttons (still restricted)
- [x] With `rbac__project_roles` flag enabled:
  - Verify all existing functionality works unchanged

### DIFF
--- a/front/lib/front_web/templates/people/members/_member.html.eex
+++ b/front/lib/front_web/templates/people/members/_member.html.eex
@@ -26,12 +26,8 @@
     </div>
 
     <% member_role_names = Enum.map(@member.subject_role_bindings, &(&1.role.name)) %>
-    <% show_buttons? = if @member_type == :service_account do
-      show_service_account_management_buttons?(@conn, @org_scope?, @permissions)
-    else
-      show_people_management_buttons?(@conn, @org_scope?, @permissions)
-    end %>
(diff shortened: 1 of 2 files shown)


front/lib/front_web/templates/people/members/_member.html.eex
```diff
@@ -26,8 +26,12 @@
     </div>
 
     <% member_role_names = Enum.map(@member.subject_role_bindings, &(&1.role.name)) %>
-    <%= if show_people_management_buttons?(@conn, @org_scope?, @permissions)
-      && "Owner" not in member_role_names do %>
+    <% show_buttons? = if @member_type == :service_account do
```
skipi: Can we extract this to a helper?

skipi (approved):

PR #768, opened 2025-12-16
Feature/create deletion workers

loadez:
## ✅ Checklist


artifacthub/cmd/server/main.go
```diff
@@ -159,6 +170,10 @@ func main() {
 		go bucketcleanerWorker(storageClient)
 	}
 
+	if os.Getenv("START_JOBDELETION_WORKER") == "yes" {
```
skipi: typo

loadez: Fixed

skipi (dismissed):

zebra/lib/zebra/workers/job_deletion_policy_worker.ex
```diff
@@ -67,12 +67,26 @@ defmodule Zebra.Workers.JobDeletionPolicyWorker do
...
+        Enum.map(jobs_expired, fn {id, _organization_id, _project_id, _artifact_store_id} ->
+          id
+        end)
+
+      Logger.info("Found #{length(job_ids)} expired jobs for deletion: #{inspect(job_ids)}")
+
+      with :ok <- Job.publish_job_deletion_events(jobs_expired) do
+        {:ok, deleted_stop_requests} = Job.delete_job_stop_requests(job_ids)
+        {:ok, deleted_jobs} = Job.delete_jobs(job_ids)
+        {:ok, deleted_stop_requests, deleted_jobs}
```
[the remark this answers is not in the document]
loadez: No problem here, loghub and artifacthub tolerate duplicated requests fine.

hamir-suspect (dismissed):

artifacthub/pkg/workers/jobdeletion/worker.go
```diff
@@ -0,0 +1,111 @@
+package jobdeletion
```
dexyk, line 1: @loadez please add the tests for the worker, thanks

loghub2/pkg/workers/jobdeletion/worker.go
```diff
@@ -0,0 +1,104 @@
+package jobdeletion
```
dexyk, line 1: @loadez please add the tests for the worker, thanks

artifacthub/pkg/workers/jobdeletion/worker.go
```diff
@@ -0,0 +1,111 @@
...
+	event := &server_farm_job.JobDeleted{}
+
+	err := proto.Unmarshal(delivery.Body(), event)
+	if err != nil {
+		log.Printf("JobDeletion Worker: Failed to parse message: %s, error: %+v", delivery.Body(), err)
+		return err
+	}
+
+	jobID := event.GetJobId()
+	artifactStoreID := event.GetArtifactStoreId()
```
dexyk, line 88: @loadez if artifact store ID is empty we should skip the processing and just log the case

zebra/lib/zebra/models/job.ex
```diff
@@ -400,42 +400,43 @@ defmodule Zebra.Models.Job do
...
+  def expired_job_ids(limit) do
     import Ecto.Query,
-      only: [from: 2, where: 3, subquery: 1, limit: 2, order_by: 2]
+      only: [from: 2, join: 5, where: 3, limit: 2, select: 2]
 
-    jobs_subquery =
+    query =
       from(j in Zebra.Models.Job,
+        join: p in "projects",
+        on: j.project_id == p.id,
```
dexyk, line 410: @loadez what if project was deleted

zebra/lib/zebra/models/job.ex
```diff
@@ -609,6 +610,82 @@ defmodule Zebra.Models.Job do
...
+        {:error, "AMQP channel unavailable"}
+    end
+  rescue
+    error ->
+      Logger.error("Exception in publish_job_deletion_events: #{inspect(error)}")
+      {:error, "Publishing failed with exception"}
+  end
+
+  defp publish_single_job_deletion(channel, exchange_name, routing_key, job_meta) do
+    {id, organization_id, project_id, artifact_store_id} = job_meta
```
dexyk, line 646: @loadez what if artifact_store_id is nil

zebra/lib/zebra/workers/job_deletion_policy_worker.ex
```diff
@@ -67,12 +67,26 @@ defmodule Zebra.Workers.JobDeletionPolicyWorker do
   end
```
dexyk, line 67: @loadez please add the tests for the worker, thanks

dexyk (changes requested):

zebra/lib/zebra/workers/job_deletion_policy_worker.ex
```diff
@@ -67,12 +67,26 @@ defmodule Zebra.Workers.JobDeletionPolicyWorker do
...
+         true <- length(jobs_expired) > 0 do
+      job_ids =
+        Enum.map(jobs_expired, fn {id, _organization_id, _project_id, _artifact_store_id} ->
+          id
+        end)
+
+      Logger.info("Found #{length(job_ids)} expired jobs for deletion: #{inspect(job_ids)}")
+
+      with :ok <- Job.publish_job_deletion_events(jobs_expired) do
+        {:ok, deleted_stop_requests} = Job.delete_job_stop_requests(job_ids)
```
dexyk: @loadez are these using `=` intentionally instead of `<-` for pattern matching?

dexyk (changes requested):

zebra/lib/zebra/models/job.ex
loadez: We should still emit the message, it will end up in a dead letter on the artifact hub. In a new pr I will add a worker on project hub that will receive the messages from zebra and we can handle that there, but if we don't know the artifact store of a job I think there is nothing we can do from the services.

zebra/lib/zebra/models/job.ex
loadez: If the project was deleted that job won't be returned in this query. We didn't talk about the case where the project was deleted but because of the query being inner it will naturally skip those records and we can implement a solution to this case later.

zebra/lib/zebra/workers/job_deletion_policy_worker.ex
loadez: No, this was an error.

zebra/lib/zebra/models/job.ex
dexyk: @loadez ok, but wouldn't `Ecto.UUID.load(artifact_store_id)` fail in that case?

loadez: Now this is handled on lines 648-652

artifacthub/pkg/workers/jobdeletion/worker.go
```diff
@@ -0,0 +1,116 @@
...
+	log.Printf("JobDeletion Worker: Starting consumer for exchange=%s routing_key=%s", JobDeletionExchange, JobDeletionRoutingKey)
+	go func() {
+		err := w.consumer.Start(w.amqpOptions, w.handleMessage)
+		if err != nil {
+			log.Printf("JobDeletion Worker: error starting consumer %s", err)
+
+			w.reconnectAttempts++
+			waitTime := max(w.reconnectAttempts*2, 60)
+			time.Sleep(time.Duration(waitTime) * time.Second)
+			w.Start()
```
dexyk: @loadez please use a single loop here inside `go func` instead of recursion

loghub2/pkg/workers/jobdeletion/worker.go
```diff
@@ -0,0 +1,106 @@
...
+	log.Printf("JobDeletion Worker: Starting consumer for exchange=%s routing_key=%s", JobDeletionExchange, JobDeletionRoutingKey)
+	go func() {
+		err := w.consumer.Start(w.amqpOptions, w.handleMessage)
+		if err != nil {
+			log.Printf("JobDeletion Worker: error starting consumer %s", err)
+
+			w.reconnectAttempts++
+			waitTime := max(w.reconnectAttempts*2, 60)
+			time.Sleep(time.Duration(waitTime) * time.Second)
+			w.Start()
```
dexyk: @loadez please use a single loop here inside `go func` instead of recursion

artifacthub/pkg/workers/jobdeletion/worker.go
loadez: Fixed

loghub2/pkg/workers/jobdeletion/worker.go
loadez: Fixed

artifacthub/pkg/workers/jobdeletion/worker.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (w *Worker) Start() {
+	log.Printf("JobDeletion Worker: Starting consumer for exchange=%s routing_key=%s", JobDeletionExchange, JobDeletionRoutingKey)
+	go func() {
+		for {
+			err := w.consumer.Start(w.amqpOptions, w.handleMessage)
+			if err != nil {
+				log.Printf("JobDeletion Worker: error starting consumer %s", err)
+				w.reconnectAttempts++
+				waitTime := max(w.reconnectAttempts*2, 60)
```
dexyk: @loadez should we use `min` here?

loghub2/pkg/workers/jobdeletion/worker.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (w *Worker) Start() {
+	log.Printf("JobDeletion Worker: Starting consumer for exchange=%s routing_key=%s", JobDeletionExchange, JobDeletionRoutingKey)
+	go func() {
+		for {
+			err := w.consumer.Start(w.amqpOptions, w.handleMessage)
+			if err != nil {
+				log.Printf("JobDeletion Worker: error starting consumer %s", err)
+				w.reconnectAttempts++
+				waitTime := max(w.reconnectAttempts*2, 60)
```
dexyk: @loadez should we use `min` here?

dexyk (dismissed):

dexyk (approved):

zebra/lib/zebra/models/job.ex
```diff
@@ -609,6 +610,86 @@ defmodule Zebra.Models.Job do
...
+
+    artifact_store_id_str =
+      case Ecto.UUID.load(artifact_store_id) do
+        {:ok, uuid_str} -> uuid_str
+        :error -> ""
+      end
```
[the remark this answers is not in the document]
loadez: Tackle will handle this case after a couple of retries.

So it won't try this forever.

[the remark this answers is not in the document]
skipi: It'll also land on a dead-letter, so we can debug it on a per-case basis.

PR #769, opened 2025-12-16
docs: update README installation links for CE

TomFern:
## 📝 Description
Update Semaphore CE installation links in the README

## ✅ Checklist


radwo (approved):

README.md
```diff
@@ -29,20 +29,22 @@ ___
 ___
 
 ## Installation
+
 Installing and running Semaphore is easy and only takes **10-30 minutes** ⏱️
 
 Choose your preferred installation method from our detailed guides:
 
 ### 🎡 Kubernetes Cluster
```
radwo: We have now only 3 links, so maybe put them in row and name them

`[Kubernetes Cluser][Single Machine][Local Machine]`

What do you think? in that case we won't need to revisit the docs when we will add another cluster or antoher provider for a single machine.

radwo (changes requested):

TomFern:
@radwo I updated the badges and links

PR #813, opened 2026-01-16
feat(notifications): add retry logic with exponential backoff for webhook timeouts

dexyk:
## 📝 Description

Adds retry logic with exponential backoff for failed webhook deliveries, improving reliability.

## ✅ Checklist


DamjanBecirovic (approved):

loadez (commented):
@dexyk shouldn't we put a cap on the delay too along with a max delay? If someone want's to change the max retries to 10 for example they may not notice that this would mean that we have a delay of almost 6 days. We can define it as a 60s max

loadez (approved):

PR #829, opened 2026-01-27
fix(zebra): remove join from zebra retention queries

skipi:
## 📝 Description
Split the query into two separate queries - first fetch expired job IDs using the index, then lookup artifact_store_id from projects table - to avoid slow heap access on a table with a lot of dead tuples.

## ✅ Checklist

### DIFF
--- a/zebra/lib/zebra/models/job.ex
+++ b/zebra/lib/zebra/models/job.ex
@@ -402,18 +402,35 @@ defmodule Zebra.Models.Job do
 
   def expired_job_ids(limit) do
     import Ecto.Query,
-      only: [from: 2, join: 5, where: 3, limit: 2, select: 2]
+      only: [from: 2, where: 3, limit: 2, select: 2, order_by: 2]
 
-    query =
+    jobs_query =
       from(j in Zebra.Models.Job,
-        join: p in "projects",
-        on: j.project_id == p.id,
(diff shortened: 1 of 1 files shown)


zebra/lib/zebra/models/job.ex
```diff
@@ -402,18 +402,35 @@ defmodule Zebra.Models.Job do
...
-      only: [from: 2, join: 5, where: 3, limit: 2, select: 2]
+      only: [from: 2, where: 3, limit: 2, select: 2, order_by: 2]
 
-    query =
+    jobs_query =
       from(j in Zebra.Models.Job,
-        join: p in "projects",
-        on: j.project_id == p.id,
         where: not is_nil(j.expires_at) and j.expires_at <= fragment("CURRENT_TIMESTAMP"),
+        order_by: [asc: j.expires_at],
```
dexyk, line 410: @skipi do we need the order by `expires_at`? It adds some overhead and, while rare, it seems plausible that with limit applied before filtering projects we could keep picking the same orphaned jobs and return an empty result, stalling processing.

loadez: Actually using order by can force the planner to use the index when it would think that a seq scan would be cheaper/faster. This is such a case.

And I agree with the second point.

skipi: We might have to - the query planner for some reason does not want to use the index for this query. I'll move this to draft, as we have two more things we want to check before making changes in Zebra.

PR #836, opened 2026-01-29
feat(repohub): Fix inconsistent origin on repo update

loadez:
## 📝 Description

This sets the origin of the repo before updating to prevent errors on edit workflow after changing the repository name.

## ✅ Checklist
- [x] ~~This change requires documentation update~~

### DIFF
--- a/repohub/pkg/gitrekt/update_or_clone.go
+++ b/repohub/pkg/gitrekt/update_or_clone.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net/url"
 	"os/exec"
 	"strings"
 	"syscall"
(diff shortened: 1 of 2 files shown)


repohub/pkg/gitrekt/update_or_clone.go
```diff
@@ -66,6 +67,19 @@ func (o *UpdateOrCloneOperation) Update() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Minute)
 	defer cancel()
 
+	remoteURL, err := url.Parse(o.Repository.HttpURL)
```
dexyk, line 70: @loadez if it's feasable, it would be nice to have a test for the repo url change scenario

loadez: Tests where added, the current errors are unrelated

PR #846, opened 2026-02-05
feat(mcp): add scheduled tasks tools

cchristous:
## Summary

Adds three new MCP tools for managing Semaphore scheduled tasks (periodics):

- **tasks_list**: List scheduled tasks for a project with keyset pagination
- **tasks_describe**: Get task details including parameters and recent trigger history
- **tasks_run**: Trigger a task to run immediately with optional branch/parameter overrides

All three tools use the generated protobuf definitions from #916.

### Security Controls
- **Pre-RPC RBAC enforcement**: All three handlers validate permissions _before_ making any gRPC call — `project.scheduler.view` for list/describe, `project.scheduler.run_manually` for run
- **Post-RPC scope mismatch checks**: `describe` and `run` verify the returned task's `organization_id` and `project_id` match the caller-supplied values, preventing cross-tenant access. `list` performs per-item scope filtering on returned results. Uses `ReportScopeMismatch` for structured logging and metrics, matching the pattern in `jobs/describe.go`
- **Pre-flight scope verification for writes**: `run` handler calls `Describe` before `RunNow` to verify task ownership, since `RunNowRequest` only accepts a task ID without org/project scoping
- **UUID validation**: All ID parameters validated via regex patterns at schema level and handler level
- **Feature flag gating**: Read operations gated by `mcp_server_read_tools`, write operations by `mcp_server_write_tools`
- **Path traversal protection**: `pipeline_file` parameter rejects absolute paths, `..` sequences, backslashes, and control characters
- **Parameter validation**: Task parameters must start with letter/underscore, no control characters, max 128 char names
- **Audit logging**: Unauthorized run attempts are logged with full context (userId, orgId, projectId, taskId, permission)

### Implementation Safety
- Follows established patterns from existing MCP tools (jobs, pipelines, workflows)
- All proto getters use nil-safe `Get*()` methods
- Explicit nil checks on proto messages and slice elements (with warn-level logging for nil entries)
- Incomplete `RunNow` responses (nil periodic/trigger) return an error instead of silent partial success
- Context timeouts on all gRPC calls via `CallTimeout()`
- Response truncation prevents oversized outputs
- Structured logging on all RPC failures and non-OK status responses with orgId/projectId context
- Startup warning when scheduler endpoint is not configured (graceful degradation)

### Test Coverage
- 32 test cases covering:
  - Feature flag disabled scenarios (list, run)
  - Successful list/describe/run operations
  - Permission denied for all 3 handlers (verifies RPCs are NOT called, including scope-check Describe)
  - Scope mismatch checks: org and project mismatch for describe and run (verifies RunNow is NOT called)
  - List scope mismatch filtering (verifies out-of-scope tasks are excluded from results)
  - RPC error handling for all 3 handlers
  - Non-OK scheduler status for describe and list
  - Scope-check Describe failures in run handler (transport error, non-OK status, nil periodic)
  - Incomplete RunNow response handling
  - Missing X-Semaphore-User-ID header for all 3 handlers
  - Nil scheduler client for all 3 handlers
  - Run with parameters, branch, and pipeline_file overrides
  - Missing required parameters and invalid UUID validation
  - `validatePipelineFile` table-driven tests (path traversal, control chars, backslashes, absolute paths, length)
  - `validateParameterName` table-driven tests (multi-byte UTF-8, control chars, length)

## Files Changed

| File | Purpose |
|------|---------|
| `pkg/tools/tasks/list.go` | List tool — keyset pagination, summary/detailed modes, per-item scope filtering |
| `pkg/tools/tasks/describe.go` | Describe tool — task details, parameters, trigger history, post-RPC scope verification |
| `pkg/tools/tasks/run.go` | Run tool — immediate trigger with optional overrides, pre-flight scope verification via Describe |
| `pkg/tools/tasks/register.go` | Constants and tool registration |
| `pkg/tools/tasks/tasks_test.go` | 32 test cases |
| `pkg/internalapi/config.go` | Scheduler endpoint config + startup warning |
| `pkg/internalapi/manager.go` | Scheduler client in Provider interface + Manager |
| `cmd/mcp_server/main.go` | Register tasks tools |
| `test/support/stubs.go` | SchedulerClientStub with error injection |
| `test/support/mock_provider.go` | MockProvider scheduler support |
| `README.md` | Tool documentation |

## Test plan

- [x] Unit tests pass (`go test ./pkg/tools/tasks/...`) — 32/32
- [x] Full project builds (`go build ./...`)
- [ ] Manual testing with MCP client against staging environment
- [ ] Verify RBAC denies access without proper permissions
- [ ] Verify feature flags properly gate tool availability

### DIFF
--- a/docs/docs/getting-started/changelog.md
+++ b/docs/docs/getting-started/changelog.md
@@ -8,6 +8,19 @@ Thank you for using Semaphore!
 We continuously deploy changes that improve our product for our customers.
 This page is updated on a weekly basis.
 
+### Week of March 23 2026
+
+**(Improved) Ubuntu 24.04 image update**
+
+**(New)** New packages:
+
+- .NET SDK: 10.0.104
+- PowerShell: 7.6.0
+- Edge: 146.0.3856.72
+- EdgeDriver: 146.0.3856.72
+
+To learn more about this image, check our [Ubuntu 24.04 page](https://docs.semaphoreci.com/reference/os-ubuntu-images/ubuntu-2404-image/).
+
 ### Week of February 9 2026
 
 **(Improved) Ubuntu 24.04 image update**
--- a/docs/docs/reference/os-ubuntu-images/ubuntu-2404-image.md
+++ b/docs/docs/reference/os-ubuntu-images/ubuntu-2404-image.md
@@ -57,6 +57,8 @@ Following version control tools are pre-installed:
 - Geckodriver 0.36.0
 - Google Chrome 144.0.7559.59
 - ChromeDriver 144.0.7559.59
+- Microsoft Edge: 146.0.3856.72
+- EdgeDriver: 146.0.3856.72
 - Xvfb (X Virtual Framebuffer)
 - Phantomjs 2.1.1
 
@@ -218,6 +220,11 @@ The default installed Ruby version is 3.4.8.
 ### Swiftly
 
 - 1.1.1
+
+### Microsoft .NET SDK and PowerShell
+
+- .NET SDK: 10.0.104
+- PowerShell: 7.6.0
 
 ## See also
 
--- a/docs/docs/using-semaphore/languages/dotnet.md
+++ b/docs/docs/using-semaphore/languages/dotnet.md
@@ -1,0 +1,125 @@
+---
+description: .NET Guide
+sidebar_position: 9
+---
+
+# .NET
+
+Semaphore Ubuntu images include the .NET SDK and PowerShell.
+
+This page shows how to inspect installed versions, build .NET projects, and run tests.
+
+## Check installed versions
+
+Use the following commands to inspect available .NET and PowerShell versions:
+
+```bash
+dotnet --info
+dotnet --list-sdks
+dotnet --list-runtimes
+pwsh --version
+```
+
+## Build a .NET project
+
+The following example restores dependencies and builds a project in Release mode:
+
+```yaml
+version: v1.0
+name: .NET pipeline
+agent:
+  machine:
+    type: f1-standard-2
+    os_image: ubuntu2404
+blocks:
+  - name: Build
+    task:
+      jobs:
+        - name: Build application
+          commands:
+            - checkout
+            - dotnet --info
+            - dotnet restore
+            - dotnet build --configuration Release --no-restore
+```
+
+## Run tests
+
+Use `dotnet test` to execute your test suite:
+
+```yaml
+version: v1.0
+name: .NET tests
+agent:
+  machine:
+    type: f1-standard-2
+    os_image: ubuntu2404
+blocks:
+  - name: Test
+    task:
+      jobs:
+        - name: Run tests
+          commands:
+            - checkout
+            - dotnet restore
+            - dotnet build --configuration Release --no-restore
+            - dotnet test --configuration Release --no-build
+```
+
+## Use PowerShell
+
+PowerShell is available as `pwsh`.
+
+Example:
+
+```yaml
+version: v1.0
+name: .NET with PowerShell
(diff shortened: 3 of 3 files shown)


DamjanBecirovic:
@cchristous sorry for the late response here.

Most of the changes look good, but the manually created protobuf definitions file that was used here does not match the one used on the server, so it will not actually work end-to-end.
We have added to the MCP server on the main branch the protobuf definitions file that was fully generated from source definitions, so it matches the one used on the server side.

Can you please rebase this PR with the main branch and change it to use the generated protobuf definitions file instead of the original, manually created one?

mcp_server/pkg/tools/tasks/describe.go
```diff
@@ -0,0 +1,365 @@
...
+
+The authentication layer must inject the X-Semaphore-User-ID header so we can authorize task access.`, err)), nil
+		}
+
+		if err := authz.CheckProjectPermission(ctx, api, userID, orgID, projectID, schedulerViewPermission); err != nil {
+			return shared.ProjectAuthorizationError(err, orgID, projectID, schedulerViewPermission), nil
+		}
+
+		request := &schedulerpb.DescribeRequest{
+			Id: taskID,
```
DamjanBecirovic, line 172: Authz uses caller-supplied project_id, but DescribeRequest only sends task_id, and periodic.project_id / periodic.organization_id from the response are not checked against the requested scope.

This allows an attack vector where a malicious user supplies a project ID on which they have permissions, and the task_id of the scheduler from a different project or organization that they can not access. 

Please add scope mismatch checks after the describe call to solve this issue, similar to what is being done [here](https://github.com/semaphoreio/semaphore/blob/main/mcp_server/pkg/tools/jobs/describe.go#L123-L154)

mcp_server/pkg/tools/tasks/run.go
```diff
@@ -0,0 +1,405 @@
...
+			Requester:       userID,
+			Reference:       branch,
+			PipelineFile:    pipelineFile,
+			ParameterValues: pbParams,
+		}
+
+		callCtx, cancel := context.WithTimeout(ctx, api.CallTimeout())
+		defer cancel()
+
+		resp, err := client.RunNow(callCtx, request)
```
DamjanBecirovic, line 273: There is a similar issue here as in the `describe` tool implementation.
Authz uses the caller-supplied project_id, but there is no check to see if the provided task actually belongs to the project for which the user was authorized.
The implementation here should describe the task with the supplied task_id first and then check if the actual project_id/organization_id of the task matches the user-supplied values. Only once that is verified, the RunNow call can be made.

Please also add tests for these cases.

mcp_server/pkg/tools/tasks/describe.go
cchristous: Implemented. Added post-RPC scope mismatch checks in `describeHandler` that verify the returned periodic's `organization_id` and `project_id` match the caller-supplied values, using `ReportScopeMismatch` and `ScopeMismatchError` matching the pattern in `jobs/describe.go`. Also added the same check to `listHandler` for per-item filtering. Tests cover both org and project mismatch scenarios.

mcp_server/pkg/tools/tasks/run.go
cchristous: Implemented. The `runHandler` now calls `Describe` before `RunNow` to verify the task's actual `organization_id` and `project_id` match the caller-supplied values. Only after scope verification passes does the `RunNow` RPC execute. Tests added for both org and project scope mismatch (verifying `LastRunNow` is nil), plus tests for Describe failure modes during scope verification (transport error, non-OK status, nil periodic).

PR #869, opened 2026-02-18
feat(mcp): add OAuth 2.1 auth flow with Guard and Auth JWT validation

hamir-suspect:
## 📝 Description
- Adds MCP OAuth 2.1 support on mcp.<domain>, with Guard acting as the authorization server for /mcp/oauth/*.
- Routes OAuth discovery, registration, authorization, consent, and token exchange through Emissary to Guard; routes protected /mcp/* traffic to mcp-server via Auth ext_authz.
- Updates Auth to validate Guard-issued MCP JWT access tokens and set x-semaphore-user-id for downstream MCP requests.
- Preserves backward compatibility by falling back to legacy API token auth when JWT validation fails.
- Returns MCP/OAuth-compliant 401 responses (including WWW-Authenticate with resource metadata) when credentials are missing or invalid.
- Adds Guard persistence for OAuth clients and auth codes, plus periodic cleanup of expired auth codes.

## ✅ Checklist


auth/lib/auth.ex
```diff
@@ -527,6 +599,39 @@ defmodule Auth do
...
+  defp authenticate_mcp_token(token) do
+    case Auth.JWT.validate_mcp_token(token) do
+      {:ok, user_id, _claims} ->
+        {:ok, user_id}
+
+      {:error, reason} ->
+        # Fall back to legacy API token validation for existing MCP clients.
+        case authenticate_based_on_token(token) do
+          {:ok, %{id: user_id}} ->
+            Logger.info("[Auth] MCP token validated via legacy API token")
```
DamjanBecirovic: This is not very helpful and will generate a lot of logs. Let's remove it.

docs/docs/using-semaphore/ai/mcp-server.md
```diff
@@ -46,16 +46,39 @@ See [example prompts](#examples) to see a bit of what's possible.
...
+OAuth-capable clients use discovery and dynamic client registration automatically.
+
+### Connect with an API token
+
+If your client does not support OAuth yet, use an API token instead.
+
+You can obtain an API token in two ways:
+
+- [Personal API Token](../user-management#profile-token): if you don't know your personal API token, you can reset it and obtain a new one.
+- [Service Account](../service-accounts): create a service account with *Member* role and use its API token.
```
DamjanBecirovic:
```suggestion
- [Service Account](../service-accounts): create a service account with *Member* role, add it to projects that you want to access via the MCP server, and use its API token.
```

guard/lib/guard/mcp_oauth/token.ex
```diff
@@ -0,0 +1,131 @@
...
+  ## Returns
+  - `{:ok, token_response}` on success
+  - `{:error, error_response}` on failure
+  """
+  @spec exchange(map()) :: {:ok, map()} | {:error, map()}
+  def exchange(params) do
+    with :ok <- validate_grant_type(params),
+         {:ok, auth_code} <- validate_and_consume_code(params),
+         :ok <- validate_pkce(auth_code, params),
+         :ok <- validate_redirect_uri(auth_code, params),
```
DamjanBecirovic: `validate_and_consume_code/1` marks the auth code as used before PKCE (`code_verifier`) and `redirect_uri` are validated.

That means any failed token request (wrong verifier/redirect, accidental retry with bad params, or malicious request) burns a valid code and causes subsequent legitimate exchange to fail with `invalid_grant`.

Could we change this flow to:
1) load + lock valid unused code (`FOR UPDATE`) in a transaction,
2) validate PKCE + redirect_uri,
3) set `used_at` only after validation passes?

This keeps one-time use semantics and avoids pre-validation code invalidation.

DamjanBecirovic (changes requested):
There are no tests included for the new behavior. 

At a minimum, we need tests for:
- in `auth` happy path + plus scenarios with missing or invalid token
- in `guard` verifying behavior for obtaining and locking the auth code
- in `guard` tests for `mcp_oauth/server.ex` asserting that it returns what is expected for supported endpoints
- in `guard` test for deleting auth_codes

DamjanBecirovic (dismissed):

DamjanBecirovic (approved):

DamjanBecirovic (dismissed):

guard/lib/guard/mcp_oauth/server.ex
```diff
@@ -0,0 +1,478 @@
...
+            Authorize.build_error_redirect(
+              redirect_uri,
+              "invalid_request",
+              "Invalid redirect_uri",
+              state
+            )
+
+          conn
+          |> put_resp_header("location", error_url)
+          |> send_resp(302, "")
```
dexyk: @hamir-suspect shouldn't we return 400 instead of redirecting to the invalid URI?

dexyk (changes requested):

guard/lib/guard/mcp_oauth/server.ex
hamir-suspect: good catch thanks

dexyk (dismissed):

dexyk (dismissed):

DamjanBecirovic (approved):

dexyk (approved):

skipi (approved):

PR #873, opened 2026-02-20
feat(github_hooks): migrate installation repositories to dedicated table

radwo:
- add github_app_installation_repositories with installation_id, slug, and stored remote_id
- backfill existing installation repositories into the new table
- switch repository/org lookups to table-backed queries
- keep add/remove/refresh flow slug-based with case-insensitive matching
- remove remote-id update path from refresh
- sync legacy github_app_installations.repositories via callbacks during rollout
- add specs for refresh diffing and slug-based matching behavior

## ✅ Checklist

### DIFF
--- a/github_hooks/Gemfile
+++ b/github_hooks/Gemfile
@@ -33,7 +33,7 @@ gem "sentry-rails"
 gem "sentry-sidekiq"
 
 gem "sprockets", "< 4"
-gem "nokogiri", "~> 1.18.9"
+gem "nokogiri", "~> 1.19.0"
 gem "uri", "~> 1.0", ">= 1.0.4"
 
 gem "rt-watchman", :require => "watchman", :github => "renderedtext/watchman", :ref => "74530687a232aea678b6738114c82dfc163657cd"
--- a/github_hooks/Gemfile.lock
+++ b/github_hooks/Gemfile.lock
@@ -244,15 +244,15 @@ GEM
     net-smtp (0.5.1)
       net-protocol
     nio4r (2.7.4)
-    nokogiri (1.18.9-aarch64-linux-gnu)
+    nokogiri (1.19.1-aarch64-linux-gnu)
       racc (~> 1.4)
-    nokogiri (1.18.9-arm-linux-gnu)
+    nokogiri (1.19.1-arm-linux-gnu)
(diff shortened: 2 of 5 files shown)


radwo:
@codex review

github_hooks/db/migrate/20260220101000_create_github_app_installation_repositories.rb
```diff
@@ -0,0 +1,54 @@
...
+  def up
+    create_table :github_app_installation_repositories, id: :uuid, default: -> { "uuid_generate_v4()" } do |t|
+      t.bigint :installation_id, null: false
+      t.bigint :remote_id, null: false, default: 0
+      t.string :slug, null: false
+
+      t.timestamps null: false
+    end
+
+    execute <<~SQL
```
skipi, line 15: 🤔 I'm wondering if that's the right way of doing this:
- We're locking the table for the time of the indexes being created(I'm not sure what data volume we're talking about here)
- the query for seeding the data might also take some time(given the indexes are created afterwards)

I'd:
- separate seed/index creation(if possible)
- make index creation concurrent(`create index concurrently`)
- run seed script manualy

github_hooks/lib/semaphore/github_app/repositories/remote_id_backfill.rb
```diff
@@ -0,0 +1,216 @@
+module Semaphore::GithubApp
+  class Repositories
+    class RemoteIdBackfill
+      MAX_NUMBER_OF_REPOSITORIES = Semaphore::GithubApp::Repositories::MAX_NUMBER_OF_REPOSITORIES
+      QUERY_BATCH_SIZE = 500
+      ADVISORY_LOCK_NAMESPACE = 71_104
```
skipi, line 8: Can we add a note on how this number was chosen?

github_hooks/db/migrate/20260220101000_create_github_app_installation_repositories.rb
radwo: We discuss this in DMs. 

For summary: There will be only ~133k records in this table. This table will be used only after the migration is complete.

skipi (approved):

PR #893, opened 2026-02-28
fix(repository_hub): use direct reference lookup for GitHub tag resolution

cchristous:
## Summary

- Replace `Tentacat.Repositories.Tags.list` (lists ALL tags, paginated at 30/page) with `Tentacat.References.find` (single O(1) direct lookup via `GET /repos/{owner}/{repo}/git/refs/tags/{tag_name}`)
- For annotated tags, dereference the tag object to its commit SHA with one additional call to `GET /repos/{owner}/{repo}/git/tags/{sha}`
- Add catch-all clause for unexpected object types to prevent `FunctionClauseError`
- Add test coverage for annotated tag dereference path

## Why this is the right fix

The `get_tag` function listed ALL repository tags from GitHub's API and searched linearly. For a repo with 500 tags, this required ~17 paginated API calls per lookup. During bursts of pact webhook retries (226 in the worst hour), this amplified to ~3,800 GitHub API calls just for tag lookups, exhausting rate limits and causing cascading "Tag does not exist" failures that contributed to an incident.

GitLab and Bitbucket adapters already use direct single-tag endpoints. The GitHub adapter was the outlier. The Tentacat library already ships `Tentacat.References.find/4` which does exactly what we need — it just wasn't being used.

## Why this is safe

- **Return contract unchanged**: `get_tag` still returns `{:ok, %{type: "tag", sha: <commit_sha>}}`. No caller changes needed.
- **Error semantics preserved**: 404 → `:not_found`, other errors → `:precondition` with logging.
- **Annotated tags handled explicitly**: `resolve_tag_commit/4` pattern-matches on `object.type` to distinguish lightweight (`"commit"`) from annotated (`"tag"`) tags, with a catch-all for unexpected types.
- **All existing tests pass** plus new test coverage for the annotated tag dereference path.

## API call reduction

| Tag type    | Before  | After |
|-------------|---------|-------|
| Lightweight | ~17     | 1     |
| Annotated   | ~17     | 2     |

## Test plan

- [x] `github_client_test.exs`: 16/16 pass (including new annotated tag test)
- [x] `describe_revision_action_test.exs`: 2/2 pass (no regressions)
- [ ] CI full suite


repository_hub/lib/repository_hub/clients/github_client.ex
```diff
@@ -749,49 +749,86 @@ defmodule RepositoryHub.GithubClient do
...
+
+  defp resolve_tag_commit(payload, _client, owner, repo, tag_name) do
+    log_error([
+      "unexpected tag reference object type for #{owner}/#{repo} : #{tag_name}",
+      "object type: #{inspect(get_in(payload, ["object", "type"]))}"
+    ])
+
+    fail_with(
+      :precondition,
+      "Unexpected tag reference object type for #{owner}/#{repo} : #{tag_name}: #{inspect(get_in(payload, ["object", "type"]))}."
```
loadez: I consider this leaky, why return the `inspect(get_in(payload, ["object", "type"]))` if we already have a log for this? Can't we just return a simple error?

cchristous: Good catch — removed the inspect from the error message. The object type is already captured in the log_error call above.

loadez (approved):

skipi (approved):

PR #894, opened 2026-02-28
fix(api): return HTTP 503 for transient GitHub API failures

cchristous:
## Summary

- Transient GitHub API failures (5xx, 429) were incorrectly reported as HTTP 400 "Tag/Branch does not exist" to Task API (`run_now`) callers
- Root cause: `repository_hub` mapped **all** non-success GitHub responses (except 404/422) to gRPC `FAILED_PRECONDITION`, `periodic_scheduler` discarded the gRPC error type with a wildcard match, and `public-api` mapped `FAILED_PRECONDITION` to HTTP 400
- This fix propagates a new `UNAVAILABLE` gRPC status through the chain so transient errors become HTTP 503 instead of HTTP 400

### repository_hub (`github_client.ex`)
Split the catch-all `{status, _, response}` clause in `get_branch`, `get_tag`, and `get_commit` into two clauses:
- **5xx/429** → `fail_with(:unavailable, ...)` (gRPC status 14 — UNAVAILABLE)
- **Other errors** → `fail_with(:precondition, ...)` (unchanged behavior)

This is safe because `fail_with(:unavailable, ...)` already exists in `toolkit.ex` and the server's `handle_response` already handles the resulting status map.

Also fixed:
- Pre-existing bug where `get_commit`'s log_error referenced `params.tag_name` instead of `params.commit_sha`
- Credential leak: `get_commit` used `inspect(response)` instead of `inspect_response(response)`, potentially logging Authorization headers
- `get_tag` log messages now consistently use local bindings (`owner`/`repo`/`tag_name`) instead of mixing `params.*` access

### periodic_scheduler

**`run_now_impl.ex`**:
- `fetch_branch_revision`: Added clause matching `{:error, %GRPC.RPCError{status: @grpc_unavailable, message: message}}` → `{:error, {:unavailable, message}}` before the wildcard, preserving the UNAVAILABLE signal
- `verify_revision_exists`: Added `{:error, {:unavailable, message}}` → `ToTuple.error(:UNAVAILABLE)` so it flows through the gRPC response as status code UNAVAILABLE rather than being swallowed into a generic "Cannot find git reference" FAILED_PRECONDITION

**`schedule_wf_impl.ex`**:
- `fetch_branch_revision`: Added matching clause for `GRPC.RPCError` with UNAVAILABLE status → `{:error, {:unavailable, message}}`, consistent with `run_now_impl.ex`

**`schedule_task.ex`**:
- Added `{:error, {:unavailable, message}}` clause in `schedule_workflow/3` that triggers `restart_with_exp_backoff`, so transient GitHub failures are retried with exponential backoff instead of immediately failing the trigger

### hooks_processor (`server.ex`)
- Added `@grpc_unavailable GRPC.Status.unavailable()` module attribute for consistency with other modules
- Added specific pattern match for `{:error, %GRPC.RPCError{status: @grpc_unavailable}}` in `create_hook/2` to re-raise UNAVAILABLE errors instead of converting them to `INVALID_ARGUMENT`
- Fixed catch-all clause to use `inspect(error)` instead of bare `error` in the message

### public-api v1alpha
- `response_formatter.ex`: Added `:UNAVAILABLE` clause in all 6 response formatters (`process_apply_response`, `process_get_project_id_response`, `process_describe_response`, `process_delete_response`, `process_list_response`, `process_run_now_response`) → `{:error, {:service_unavailable, message}}`
- `common.ex`: Added `respond({:error, {:service_unavailable, message}}, ...)` → HTTP 503

### public-api v2
- `response_formatter.ex`: Added `process_response` clause for `:UNAVAILABLE` → `{:error, {:service_unavailable, message}}`
- `response.ex`: Added `respond({:error, {:service_unavailable, message}}, ...)` → HTTP 503

## Why these changes are safe

1. **No behavior change for non-transient errors** — The existing catch-all clauses remain unchanged; only 5xx/429 responses take the new path
2. **Existing infrastructure** — `fail_with(:unavailable, ...)` already exists in `toolkit.ex:164`, the gRPC server already handles the status map, and both public-api response modules already have the `respond_` helper that handles any HTTP status code
3. **New clauses are placed before catch-alls** — Elixir pattern matching ensures the new specific clauses are tried first, falling through to existing behavior for all other cases
4. **HTTP 503 is retryable** — Clients can distinguish transient failures (503) from user errors (400) and retry accordingly
5. **Retry with backoff** — The periodic scheduler's `schedule_task.ex` retries UNAVAILABLE with exponential backoff (same pattern used for `RESOURCE_EXHAUSTED`), avoiding thundering herd on GitHub outages

## Test plan

- [x] `repository_hub` github_client tests: 21 tests, 0 failures — includes tests for get_branch (500, 429), get_tag (500), get_commit (500) all returning UNAVAILABLE
- [x] `periodic_scheduler` run_now_impl tests: 17 tests, 0 failures — includes UNAVAILABLE error propagation test
- [x] `periodic_scheduler` schedule_wf_impl tests: 14 tests, 0 failures — includes UNAVAILABLE scheduling status test verifying retry behavior
- [x] `public-api/v1alpha` response_formatter tests: 31 tests, 0 failures — UNAVAILABLE tests for all 6 response formatters
- [x] `public-api/v2` schedulers_client test: UNAVAILABLE gRPC status → service_unavailable tuple
- [ ] Verify CI passes on all affected components


loadez (dismissed):
The failed pipeline is due to a problem already present on main

adbatista:
@cchristous, some tests are breaking. Could you take a look, please? https://semaphore.semaphoreci.com/jobs/ea3d3187-da02-4860-879f-12190b140f3f

cchristous:
Thanks for the flag @adbatista — fixed and pushed.

The two `RepositoryHub.GithubClientTest` failures (`get_tag returns unavailable on GitHub 500`/`429`) were a test-setup bug, not a regression in the fix. Both tests mocked `Tentacat.Repositories.Tags.list`, but `get_tag` calls `Tentacat.References.find/4`, so the override never took effect — the base stub's happy-path `200` was returned and `get_tag` resolved to `{:ok, ...}`. Switched the mocks to `Tentacat.References.find/4`, matching the `get_branch` unavailable tests.

Verified locally (OTP 25, matching CI): `26 tests, 0 failures`.

The MixAudit "Security Toolbox Policy Failures" (grpc/protobuf advisories) are unrelated to this PR — it doesn't touch any `mix.lock`/`mix.exs`, so those are pre-existing dependency advisories on `main`.

adbatista:
Thank you, I'll address the mix audit ones myself 😄

PR #904, opened 2026-03-04
Fix/gitlab missing semaphore yml

loadez:
## ✅ Checklist

### DIFF
--- a/front/lib/front/models/project.ex
+++ b/front/lib/front/models/project.ex
@@ -261,10 +261,12 @@ defmodule Front.Models.Project do
     alias InternalApi.Repository.GetFileRequest
     alias InternalApi.Repository.RepositoryService.Stub
 
+    # commit_sha also accepts branch names or empty string — each git provider
+    # handles the empty case by defaulting to the repository's default branch.
     req =
       GetFileRequest.new(
         repository_id: project.repo_id,
-        commit_sha: default_branch(project),
+        commit_sha: project.repo_default_branch || "",
         path: path
       )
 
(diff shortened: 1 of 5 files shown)


front/lib/front/models/project.ex
```diff
@@ -256,29 +256,39 @@ defmodule Front.Models.Project do
...
+          "Looking for a file #{path} in project #{project.id} failed: #{inspect(error)}"
         )
 
         false
     end
   end
 
+  defp default_branch(%__MODULE__{repo_default_branch: branch})
+       when branch in [nil, ""],
+       do: "main"
```
DamjanBecirovic: It would be better to pass an empty string here and handle this case on the repository_hub side in a git-provider-specific way.

GitHub will default to the repository's default branch in case a reference is not passed, so that will solve the `master` vs `main` issue that can happen here.
According to [docs](https://docs.gitlab.com/api/repository_files/) for GitLab, the same behavior can be achieved if we use `HEAD` as a reference value.

front/lib/front/models/project.ex
```diff
@@ -256,29 +256,39 @@ defmodule Front.Models.Project do
...
-    req = GetFileRequest.new(project_id: project_id, path: path)
+  @spec file_exists?(t(), String.t()) :: boolean
+  def file_exists?(project = %__MODULE__{}, path) do
+    alias InternalApi.Repository.GetFileRequest
+    alias InternalApi.Repository.RepositoryService.Stub
+
+    req =
+      GetFileRequest.new(
+        repository_id: project.repo_id,
+        commit_sha: default_branch(project),
```
DamjanBecirovic: Please add a comment here explaining why both branch and commit_sha can be used here, since it looks like a mistake at first glance.

PR #961, opened 2026-03-27
feat: extend retention policies

skipi:
## 📝 Description
- **Plumber now emits deletion events** - when pipelines expire and are deleted, plumber publishes `PipelineDeleted` and `WorkflowDeleted` events to RabbitMQ. Workflow deletion is only emitted once all pipelines in a workflow are gone.
- **Artifacthub gains two new workers** - a `pipelinedeletion` and `workflowdeletion` worker that consume those events and clean up the associated artifacts in the bucket.
- **Plumber's retention internals were refactored** - the old policy consumer/record deleter was split into two explicit workers (policy marker and deleter) with cleaner separation of concerns, replacing the previous `policy_applier`/`record_deleter` structure.

## ✅ Checklist


plumber/ppl/test/retention/deleter/queries_test.exs
```diff
@@ -0,0 +1,278 @@
...
+
+    with_mock Ppl.Retention.Events,
+      publish_pipeline_deleted: fn _, _, _, _, _ -> :ok end,
+      publish_workflow_deleted: fn _, _, _, _ -> :ok end do
+      {:ok, 1} = Queries.delete_expired_batch(100)
+    end
+
+    assert {:error, _} = PplRequestsQueries.get_by_id(ppl_id)
+    assert {:error, _} = Block.describe(ppl_blk_0.block_id)
+    assert {:error, _} = Block.describe(ppl_blk_1.block_id)
```
DamjanBecirovic, line 229: Let's just add here an assertion that the LatestWorkflows record was also deleted.

DamjanBecirovic (approved):

PR #979, opened 2026-04-14
feat(mcp_server): add tools for listing and generating signed URLs for artifacts

dexyk:
## 📝 Description

This PR adds MCP server support for basic artifact operations: listing artifacts and retrieving signed URLs to obtain artifact files. It enforces the required project permissions for both operations 

## ✅ Checklist


artifacthub/pkg/api/public/publicapi.go
```diff
@@ -117,23 +121,36 @@ func generateSignedURLsList(ctx context.Context, client storage.Client, artifact
...
 			if err != nil {
 				return err
 			}
 
-			urls = append(urls, url)
+			attemptURLs = append(attemptURLs, url)
+		}
+
+		if !limitExceeded {
+			urls = attemptURLs
```
loadez: Correct me if I'm wrong but limitedExceeded starts as false then is set to true inside of the for if we exceed the limit and right after it exits, so if we reach the if at line 137 limitExceeded is always false and you could remove it

artifacthub/pkg/api/private/privateapi.go
```diff
@@ -121,12 +143,25 @@ func ListTransferPath(ctx context.Context, client storage.Client, artifact *mode
...
+			if limit > 0 && len(attemptResult) >= limit {
+				limitExceeded = true
+				return nil
+			}
+
+			attemptResult = append(attemptResult, &artifacthub.ListItem{Name: item.Path, IsDirectory: item.IsDirectory, Size: item.Size})
+		}
+
+		if !limitExceeded {
+			result = attemptResult
```
loadez: Same thing as the other limitExceeded if

loadez (dismissed):

artifacthub/pkg/api/public/publicapi.go
dexyk: @loadez those are inside of `err = retry.OnFailure(ctx, "Listing Bucket path", func() error {` function and the value is later used outside of the function `if limitExceeded {   return nil, ErrTooManyArtifacts }`

artifacthub/pkg/api/private/privateapi.go
dexyk: @loadez those are inside of `err = retry.OnFailure(ctx, "Listing Bucket path", func() error {` function and the value is later used outside of the function `if limitExceeded {   return nil, ErrTooManyArtifacts }`

loadez (dismissed):

mcp_server/pkg/tools/artifacts/list.go
```diff
@@ -0,0 +1,247 @@
...
+2. List artifacts under a subdirectory:
+   artifacts_list(organization_id="...", scope="workflows", scope_id="...", path="test-results")
+
+3. List project artifacts:
+   artifacts_list(organization_id="...", scope="projects", scope_id="...")
+
+Typical workflow:
+1. Call artifacts_list(...) to discover available paths.
+2. Call artifacts_signed_url(...) with the selected file path to retrieve a signed URL.`
+}
```
DamjanBecirovic, line 77: Let's add here something like the line below, since agents were struggling with that last step when signed URLs were returned by the job logs tool for self-hosted jobs.

```suggestion
3. Use the returned signed URL to download the requested file locally.
}
```

loadez (approved):

DamjanBecirovic (approved):

PR #998, opened 2026-04-23
fix(guard): store token_expires_at in Bitbucket OAuth callback

loadez:
## 📝 Description

Fixes Bitbucket OAuth token lifecycle so refresh works correctly.

**Changes:**
- **Persist `token_expires_at`** from the OAuth callback (`Guard.Id.Api.extract_repo_host_data/1`) and add it to the `RepoHostAccount` changeset cast/whitelist plus `drop_if_skip_credentials`. Without this, the column was never written, so the refresh path could not tell when a token had expired.
- **Switch `validate_token/1` to `Authorization: Bearer` header** instead of the deprecated `?access_token=` query param. Avoids leaking tokens into URL/access logs and aligns with current Bitbucket API guidance.
- **Improve refresh error logging**: include HTTP status, response body, and `repo_host_account.id` for both 4xx (revoked) and unexpected status branches; tag the missing-refresh-token warning with the account id; replace the deprecated `Logger.warn/1` with `Logger.warning/2`.
- Test mocks updated to drop the query-param URL form.

**Why:** users on Bitbucket were silently losing access because Guard never stored `token_expires_at`, so the proactive refresh check never fired and revoked accounts were hard to diagnose from logs.

## ✅ Checklist


skipi (commented):
## Code Review (gsd-code-reviewer, standard depth)

Posted as a log of findings — not a formal change request. **11 findings**: 2 BLOCKER · 5 WARN · 4 INFO.

### 🚫 Blockers

#### CR-01 — `DateTime.from_unix!/2` raises on bad input, crashes the OAuth callback
`guard/lib/guard/id/api.ex:120-122`

`extract_repo_host_data/1` uses the bang variant on provider-controlled `auth.credentials.expires_at`. An out-of-range integer raises `ArgumentError` inside the Plug pipeline → 500 response, never reaches the existing error-redirect path. Same code path runs for github / bitbucket / gitlab, so a single misbehaving upstream DoSes the callback.

```elixir
expires_at when is_integer(expires_at) ->
  case DateTime.from_unix(expires_at, :second) do
    {:ok, dt} -> dt
    {:error, reason} ->
      Logger.warning("Invalid expires_at from OAuth: #{inspect(expires_at)} (#{inspect(reason)})")
      nil
  end
```

#### CR-02 — Zero positive test for the `token_expires_at` persistence path
`guard/test/guard/api/bitbucket_test.exs`, `guard/test/guard/grpc_servers/user_server_test.exs`

The PR's stated purpose is "store `token_expires_at` in Bitbucket OAuth callback" but no test asserts it's persisted:
- `bitbucket_test.exs:67` only asserts `updated_rha.token == "new_token"` — never reads `.token_expires_at`.
- `user_server_test.exs` changes are URL-string swaps in mocks.
- No unit test on `extract_repo_host_data/1` covering nil / integer / `%DateTime{}` / malformed branches.

Future regressions in cast / whitelist / `drop_if_skip_credentials` wiring (the primary subject of this PR) won't be caught by CI. Suggested addition:

```elixir
test "refresh persists token_expires_at" do
  before = DateTime.utc_now()
  assert {:ok, {"new_token", _}} = Bitbucket.user_token(rha)
  updated = Guard.FrontRepo.get!(Guard.FrontRepo.RepoHostAccount, rha.id)
  assert updated.token_expires_at != nil
  assert DateTime.diff(updated.token_expires_at, before) in 3590..3610
end
```

Plus a unit test on `extract_repo_host_data/1` per branch.

---

### ⚠️ Warnings

**WR-01** — `inspect(body)` of OAuth refresh failure may leak echoed creds (`guard/lib/guard/api/bitbucket.ex:80, 87`). OAuth servers can echo the request `refresh_token` in `error_description`. Log only `body["error"]` and `body["error_description"]`.

**WR-02** — `:token_expires_at` is in the cast whitelist + `drop_if_skip_credentials`, but missing from `drop_if_same` in `filter_update_data` (`repo_host_account.ex:269-281`). Every callback now dirties the row + emits a "Successfully updated" log even when nothing changed. Add `|> drop_if_same(:token_expires_at, account.token_expires_at)`.

**WR-03** — Stray `validate_token` mock in the refresh test (`bitbucket_test.exs:54-58`) — the URL is never invoked in this test path; masks future regressions where `user_token` accidentally calls `validate_token`. Drop the unused mock clause.

**WR-04** — `@type t` in `RepoHostAccount` not updated — schema gained `field(:token_expires_at, :utc_datetime)` but the typespec was not (`repo_host_account.ex:24-36`). Add `token_expires_at: DateTime.t() | nil`.

**WR-05** — `extract_repo_host_data/1` clauses cover `nil | integer | %DateTime{} | other` — float and string `expires_at` fall to `other → nil`, then `OAuth.valid_token?(... nil_valid: false)` forces refresh on every request. For Bitbucket specifically this is unlikely (strategy returns integer seconds-from-epoch), but the function is shared with github/gitlab. Either handle floats explicitly or document the accepted shapes.

---

### ℹ️ Info

- **IN-01** — `update_account/2` `Logger.info "... #{inspect(account)} to #{inspect(data)}"` (`repo_host_account.ex:333`) leaks `token`, `refresh_token`, now `token_expires_at`. Pre-existing but the PR makes it more frequent (see WR-02). Redact via `inspect(Map.drop(Map.from_struct(account), [:token, :refresh_token]))`.
- **IN-02** — `Logger.warn/1` → `Logger.warning(...)` is correct, but call sites pass a single string (so it's `Logger.warning/1`, not `/2` with metadata). Consider `Logger.warning(msg, repo_host_account_id: repo_host_account.id)` for filterable logs.
- **IN-03** — Verify `RepoHostAccount.update_token/4` still has callers; otherwise remove.
- **IN-04** — Bearer-header switch in `validate_token/1` is correct. Heads-up: 401-vs-403 status surface differs slightly on bad tokens; current `200..299` check is fine.

---

### ✅ What looks good

- `:token_expires_at` correctly threaded through `create/1`, `update_account/2`, `reset_account/3` cast lists + `drop_if_skip_credentials/2`.
- The `?access_token=` query-string token-leak surface in `validate_token/1` is fully removed.
- Adding `repo_host_account.id` to error logs (without echoing tokens) is the right hygiene.

---

_Reviewer: Claude (gsd-code-reviewer), depth=standard. Rev: f755d3f._

skipi:
Addressing now (customer blocked, keeping scope tight):

- **CR-01** — switch `DateTime.from_unix!/2` → `from_unix/2` with nil fallback in `extract_repo_host_data/1`
- **WR-01** — log only `error` / `error_description` from OAuth refresh failure body, not `inspect(body)`
- **CR-02 (minimal)** — single regression test on `extract_repo_host_data/1` so the bad-integer branch can't crash again

Deferring to follow-up: WR-02 (drop_if_same), WR-03 (stray mock), WR-04 (typespec), WR-05 (float/string expires_at), all INFOs (IN-01 is pre-existing log redaction, separate ticket).

skipi:
Addressed in 1db604bf:

- ✅ **CR-01** — `extract_repo_host_data/1` no longer crashes on bad `expires_at`. Refactored into `Guard.Id.Api.parse_expires_at/1` using non-bang `DateTime.from_unix/2` with nil fallback + warning log on out-of-range.
- ✅ **WR-01** — refresh-token failure logs in `Guard.Api.Bitbucket.fetch_token` now log only `body["error"]` / `body["error_description"]`, not `inspect(body)`. Helpers `safe_oauth_error/1` + `safe_oauth_error_description/1` guard non-map bodies.
- ✅ **CR-02 (minimal)** — 5 regression tests on `parse_expires_at/1` covering nil / integer / `%DateTime{}` / out-of-range integer (the previous crash path) / unexpected shapes (string, float).

Local run: `make test.ex FILE="test/guard/id/api_test.exs test/guard/api/bitbucket_test.exs"` → **48 tests, 0 failures**.

dexyk (approved):

PR #1001, opened 2026-04-24
feat: improve error handling in scheduler and projecthub GRPC server

dexyk:
## 📝 Description

Fix Projecthub detailed serialization to return upstream scheduler/task list errors instead of crashing. Preserve Run Now failure details through periodic-scheduler and Front so users see the actual upstream error in the UI.

## ✅ Checklist


projecthub/lib/projecthub/api/grpc_server.ex
```diff
@@ -212,10 +224,16 @@ defmodule Projecthub.Api.GrpcServer do
...
-        )
+        case serialize(project) do
+          {:ok, serialized_project} ->
+            ForkAndCreateResponse.new(
+              metadata: status_ok(req),
+              project: serialized_project
+            )
+
+          {:error, reason} ->
+            error_create_response(req, serialize_error_message(reason))
```
skipi: I think this should return `ForkAndCreateResponse` - now it's returning `CreateResponse`

skipi (changes requested):

skipi (dismissed):

skipi (approved):

PR #1004, opened 2026-04-29
feat: add audit logging for artifact and rebuild operations in v1alpha API and MCP server

dexyk:
## 📝 Description

Adds audit logging in `public-api/v1alpha` and MCP server for artifact operations and workflow rebuilds.  
Aligns these endpoints with existing audit behavior in other services and enforces audit checks on critical paths. 

## ✅ Checklist


skipi:
**Inconsistent feature-check failure handling between Go and Elixir paths**

In the MCP server, when `shared.AuditLogsFeatureEnabled` returns an error (e.g. feature-hub is unreachable), the request is rejected before any audit is emitted:

- `mcp_server/pkg/tools/artifacts/signed_url.go:170-173`
- `mcp_server/pkg/tools/jobs/artifact_job_logs.go:204-207`
- `mcp_server/pkg/tools/workflows/rerun_tool.go:115-118`

```go
auditEnabled, err := shared.AuditLogsFeatureEnabled(ctx, api, orgID)
if err != nil {
    return mcp.NewToolResultError("Unable to verify audit logging availability. Please try again."), nil
}
```

The Elixir path handles the same scenario differently. In `public-api/v1alpha/lib/pipelines_api/audit.ex:160-167`, a feature-provider error is rescued, treated as "feature disabled", the stdout audit line still fires from `push_audit_log/1`, and the request proceeds:

```elixir
defp audit_logs_feature_enabled?(org_id) when is_binary(org_id) and org_id != "" do
  FeatureProvider.feature_enabled?(:audit_logs, param: org_id)
rescue
  error ->
    Logger.error("Failed to check audit_logs feature: #{inspect(error)}")
    false
end
```

So the same upstream incident (feature-hub unavailable) produces two different outcomes:

| Path | On feature-check error |
|---|---|
| v1alpha (Elixir) | stdout audit emitted, request succeeds |
| MCP server (Go) | no audit emitted at all, request rejected |

The Go behavior is strictly worse on both axes — less audit signal *and* less availability — and it's the opposite of what the Elixir module does for the same operations. Suggested fix in the Go call sites: on feature-check error, log a warning and proceed with `auditEnabled=false`. The stdout audit line in `logArtifactOperation` / `logWorkflowOperation` still fires, the operation still runs, and behavior matches the Elixir side. Something like:

```go
auditEnabled, err := shared.AuditLogsFeatureEnabled(ctx, api, orgID)
if err != nil {
    logging.ForComponent("audit").
        WithError(err).
        WithField("organization_id", orgID).
        Warn("audit_logs feature check failed; proceeding with AMQP publish disabled")
    auditEnabled = false
}
```

Worth aligning before merge so a feature-hub blip doesn't take down MCP artifact/log/rerun tools while v1alpha keeps serving.

skipi (dismissed):

skipi (approved):

PR #1011, opened 2026-05-04
allow to validate task params format

adbatista:
## 📝 Description
Adds optional regex-based input format validation to Scheduled Task parameters. When validate_input_format is enabled, a regex_pattern is enforced on default values at persist time and on submitted values at run time (both browser-side for fast feedback and server-side via PCRE match_limit for ReDoS safety). Surfaces specific param errors back to the UI, exposes the fields on public-api/v2, and round-trips them through projecthub.

## ✅ Checklist
- [X] This change requires documentation update


docs/docs/using-semaphore/tasks.md
```diff
@@ -109,6 +112,10 @@ You can add tasks by editing the project using the [Semaphore command line tool]
               - name: CANARY_VERSION
                 required: true
                 default_value: "1.0.0"
+                # Reject the task if the supplied value does not
+                # match the pattern. Pattern is capped at 512 bytes.
+                validate_input_format: true
+                regex_pattern: "^[0-9]+\\.[0-9]+\\.[0-9]+$"
```
dexyk, line 118: @adbatista we also need to update this part of the docs: https://github.com/semaphoreio/semaphore/blob/main/docs/docs/reference/project-yaml.md?plain=1#L92

dexyk (changes requested):

dexyk (approved):

PR #1014, opened 2026-05-06
feat(front): implement pagination for job listings in activity monitor

dexyk:
## 📝 Description

This PR fixes `/activity` data completeness by paginating through all pages for pipeline activity, active jobs, and debug sessions instead of reading only the first page. It also adds pagination safety guards (loop detection + max page cap of 50).

## ✅ Checklist

### DIFF
--- a/front/lib/front/activity_monitor/repo.ex
+++ b/front/lib/front/activity_monitor/repo.ex
@@ -158,27 +158,28 @@ defmodule Front.ActivityMonitor.Repo do
       url = Application.fetch_env!(:front, :pipeline_api_grpc_endpoint)
       {:ok, channel} = GRPC.Stub.connect(url)
 
-      request =
-        Req.new(
-          page_size: 300,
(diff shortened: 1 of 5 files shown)


front/lib/front/activity_monitor/repo.ex
```diff
@@ -316,9 +329,98 @@ defmodule Front.ActivityMonitor.Repo do
...
+         page_count
+       ) do
+    list_all_pages(fetch_page, next_page_token, seen_tokens, pages, page_count + 1)
+  end
+
+  defp log_pagination_stop(:max_pages, _page_token) do
+    Logger.error(
+      "Activity monitor pagination aborted: reached max page limit (max_pages=#{@max_activity_monitor_pages})"
+    )
+  end
```
loadez, line 407: nit: This should be a simple warning, unlikely to occur and the user see something still.

loadez (approved):

PR #1020, opened 2026-05-11
fix(guard): surface specific error on repo host account connect failure

adbatista:
## 📝 Description
FIX: No error is returned when connecting github account with invalid data

## ✅ Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,7 +120,7 @@ check.deps:
 		bash -c 'cd $(APP_DIRECTORY) && $(SECURITY_TOOLBOX_TMP_DIR)/dependencies --language $(LANGUAGE) -d --output-dir $(SCAN_RESULT_DIR) $(CHECK_DEPS_OPTS); EXIT_CODE=$$?; chown -R "$$HOST_UID:$$HOST_GID" $(SCAN_RESULT_DIR) 2>/dev/null || true; exit $$EXIT_CODE'
 
 check.ex.deps:
-	$(MAKE) check.deps LANGUAGE=elixir CHECK_DEPS_OPTS="-i hackney -i plug_cowboy $(CHECK_DEPS_EXTRA_OPTS)"
+	$(MAKE) check.deps LANGUAGE=elixir CHECK_DEPS_OPTS="-i hackney,plug_cowboy $(CHECK_DEPS_EXTRA_OPTS)"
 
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -120,7 +120,7 @@ check.deps:
 		bash -c 'cd $(APP_DIRECTORY) && $(SECURITY_TOOLBOX_TMP_DIR)/dependencies --language $(LANGUAGE) -d --output-dir $(SCAN_RESULT_DIR) $(CHECK_DEPS_OPTS); EXIT_CODE=$$?; chown -R "$$HOST_UID:$$HOST_GID" $(SCAN_RESULT_DIR) 2>/dev/null || true; exit $$EXIT_CODE'
 
 check.ex.deps:
-	$(MAKE) check.deps LANGUAGE=elixir CHECK_DEPS_OPTS="-i hackney -i plug_cowboy $(CHECK_DEPS_EXTRA_OPTS)"
+	$(MAKE) check.deps LANGUAGE=elixir CHECK_DEPS_OPTS="-i hackney,plug_cowboy,decimal $(CHECK_DEPS_EXTRA_OPTS)"
```
radwo: You can use now .mix-audit.txt to ignore specific CVE instead of blocking whole package

More info here https://github.com/semaphoreio/semaphore/commit/24d879ae0a1d31acf100c10c01fdac369911a0c2

front/lib/front_web/controllers/account_controller.ex
```diff
@@ -26,10 +26,59 @@ defmodule FrontWeb.AccountController do
...
+    put_flash(conn, :alert, oauth_error_text(code, provider))
+  end
+
+  defp maybe_put_oauth_flash(conn, %{"status" => "success"}) do
+    put_flash(conn, :notice, "Repository account connected.")
+  end
+
+  defp maybe_put_oauth_flash(conn, _params), do: conn
+
+  defp oauth_provider_label("github"), do: "GitHub"
```
radwo: This is necessary? Without it code will be cleaner and user won't loose much because he knows which provider he is connecting.

adbatista: 🤔 makes sense, I'll simplify, thank you

dexyk (approved):

PR #1022, opened 2026-05-12
fix(front): Don't show timer or try to fetch logs for canceled jobs

DamjanBecirovic:
## 📝 Description

Updated job and workflow UI handling for jobs stopped by fail-fast before execution starts.

Zebra already preserves the distinction via started_at == nil, so the fix was in front: stopped-before-start jobs now show --:-- instead of 00:00, the job page no longer auto-fetches logs for jobs that never ran, direct log endpoints are guarded, and the Raw logs control stays visible but disabled for those jobs. 

I also bumped Zebra’s resolved hackney dependency to a patched version to clear the CI security advisory.

## ✅ Checklist


dexyk (approved):

zebra/mix.lock
```diff
@@ -27,10 +27,10 @@
...
+  "hackney": {:hex, :hackney, "1.25.0", "390e9b83f31e5b325b9f43b76e1a785cbdb69b5b6cd4e079aa67835ded046867", [:rebar3], [{:certifi, "~> 2.15.0", [hex: :certifi, repo: "hexpm", optional: false]}, {:idna, "~> 6.1.0", [hex: :idna, repo: "hexpm", optional: false]}, {:metrics, "~> 1.0.0", [hex: :metrics, repo: "hexpm", optional: false]}, {:mimerl, "~> 1.4", [hex: :mimerl, repo: "hexpm", optional: false]}, {:parse_trans, "3.4.1", [hex: :parse_trans, repo: "hexpm", optional: false]}, {:ssl_verify_fun, "~> 1.1.0", [hex: :ssl_verify_fun, repo: "hexpm", optional: false]}, {:unicode_util_compat, "~> 0.7.1", [hex: :unicode_util_compat, repo: "hexpm", optional: false]}], "hexpm", "7209bfd75fd1f42467211ff8f59ea74d6f2a9e81cbcee95a56711ee79fd6b1d4"},
```
dexyk, line 30: @DamjanBecirovic this update could cause the issues

dexyk (changes requested):

zebra/mix.lock
DamjanBecirovic: I fixed the issue with the hackney update. The jobs are running without issue in the staging env now.
Can you take a look again please?

dexyk (approved):

PR #1025, opened 2026-05-13
feat(audit): Implement retention policy for audit events

dexyk:
## 📝 Description

Applies the system-wide 400-day data retention policy to audit logs, reusing the same `usage.apply_organization_policy` event that Zebra and Plumber consume: a `PolicyMarker` consumer stamps events older than the cutoff with an `expires_at` (clamped to a hard 400-day minimum so retention can never be shortened below policy), and a periodic `Deleter` removes expired rows using an adaptive cadence to drain backlogs.

Processing is gated per organization by the new `audit_logs_retention` feature flag, off by default and fail-closed, so retention can be rolled out gradually without risking premature deletion.

Includes the supporting schema migrations (`expires_at` column and indexes) and dependency bumps to address CVEs.

## ✅ Checklist


loadez (commented):
Two things I want to flag before we enable the deleter in prod. None of them block the merge since everything is off by default.

**1. Disabling the feature flag does not stop deletion after rows are marked.**

`check_feature_enabled` only gates the marking. If we enable `audit_logs_retention` for the wrong org, the marker sets `expires_at`, and turning the flag off later does not unmark anything. The Deleter only checks `expires_at <= now` (`deleter.ex:79` / `queries.ex:135`). So the 15 day grace is the only abort window, and the recovery is to null `expires_at` by hand in prod. Maybe re-check the org flag in the delete path? Or at least document that the flag only gates marking.

**2. The initial marking sweep is unbounded inside one AMQP message.**

`batch_mark` recurses the whole backlog of the org synchronously inside `handle_message`. The first sweep on a big org (400 days of events, batches of 1000) can pass the RabbitMQ consumer timeout, then the channel closes, the message is redelivered and the sweep restarts. With `retry_limit: 10` this can repeat a few times. It is idempotent so we don't lose anything, but maybe cap the batches per message and requeue, or pace the marking through a worker like the Deleter already does.

The rest of the safety design looks good to me. The floor clamp, the changeset excluding `expires_at`, everything off by default.

PR #1037, opened 2026-05-26
chore(zebra): improve cachehub describe observability

skipi:
## Summary

Improves visibility into the path that injects `SEMAPHORE_CACHE_*` env vars into jobs. Three previously-silent failure paths in `Zebra.Workers.JobRequestFactory.Cache.find/3` now emit warning logs and tagged metrics. The factory also logs once when cache env vars are dropped despite the project having a `cache_id` configured — one breadcrumb per impacted job.

## What changes

1. **Tagged metrics** (Watchman, positional tag, internal channel — matches `dispatcher.ex:115`)
   - `external.cachehub.describe.failed` with first tag in `grpc_error | non_ok_response | blank_credential` (unexpected, alertable)
   - `external.cachehub.describe.skipped` with first tag in `no_cache_id | forked_pr` (expected, not actionable)
2. **Logs on previously-silent paths** in `cache.ex` (non-OK status, blank credential). The gRPC error path is promoted from `Logger.info` to `Logger.warning`. The forked-PR skip stays at info — it's by design.
3. **Caller-side log** in `job_request_factory.ex`: when `cache_env_vars == []` AND project has `cache_id` AND job is not self-hosted, emit a single `Logger.warning` with `job_id`, `project_id`, `cache_id`. This is the most useful log for the operational case — "this specific job did not get cache vars".

## Behavior change worth flagging

Previously, `Cache.find(nil, _, _)` incremented `external.cachehub.describe.failed`. That was misleading: projects without a cache configured were lumped into the failure counter. This PR moves that traffic to `external.cachehub.describe.skipped` with first tag `no_cache_id`.

Any dashboard panel or alert on the old single-bucket `external.cachehub.describe.failed` series will see a drop in volume once this ships, and the skipped traffic needs to be added separately if it was being watched.

## Metric format

In SaaS (`:statsd_graphite` backend), each call emits `tagged.<prefix>.<reason>.no_tag.no_tag.external.cachehub.describe.<failed|skipped>:1|c`. Reason is the first positional tag and is filterable in Grafana via the graphite path. On-prem (`:aws_cloudwatch`) does not yet receive these metrics — see follow-up below.

## Logged fields

Logged: `cache_id`, response status code, gRPC error tuple from `inspect/1`, `job_id`, `project_id`. Not logged: cache credential, cache URL.

## Tests

- `cache_test.exs`: new `.find` describe block covering nil `cache_id`, non-OK status (asserts warning + `cache_id` in log), blank credential (asserts warning + `cache_id` + no URL leak), and the gRPC raise path (asserts warning + no credential leak).
- `job_request_factory_test.exs`: new test asserting the caller-side "Cache env vars not injected" warning fires when cachehub returns blank credential for a project with `cache_id` set.

## Follow-ups (separate work)

- Grafana panel split: stacked `cachehub.describe.failed` by reason path; alert on rate excluding `forked_pr`.
- cachehub-side per-response log for cross-checking.
- On-prem metric path (DogStatsD keyword tags via the `external:` channel) — deferred; the reported incidents come from SaaS.

### DIFF
--- a/zebra/lib/zebra/workers/job_request_factory.ex
+++ b/zebra/lib/zebra/workers/job_request_factory.ex
@@ -95,7 +95,7 @@ defmodule Zebra.Workers.JobRequestFactory do
          {:ok, open_id_token_env_vars} <-
            OpenIDConnect.load(job, repo_env_vars, organization, project, job_type, spec_env_vars),
          {:ok, callback_token} <- CallbackToken.generate(job) do
-      log_dropped_cache_vars(job, project, cache_env_vars)
+      log_dropped_cache_vars(job, project, repo_proxy, cache_env_vars)
 
       org_url = "https://#{organization.org_username}.#{Application.get_env(:zebra, :domain)}"
 
@@ -198,9 +198,11 @@ defmodule Zebra.Workers.JobRequestFactory do
     end
   end
 
-  defp log_dropped_cache_vars(job, project, cache_env_vars) do
+  defp log_dropped_cache_vars(job, project, repo_proxy, cache_env_vars) do
+    # Forked-PR cache skips are intentional (gated by the disable_forked_pr_cache
+    # feature flag in Cache.find/3), so they are not a dropped-vars condition.
     if cache_env_vars == [] and not Job.self_hosted?(job.machine_type) and
-         not is_nil(project.cache_id) do
+         not is_nil(project.cache_id) and not Cache.forked_pr?(repo_proxy) do
       Logger.warning(
         "Cache env vars not injected into job. job_id=#{job.id} project_id=#{project.id} cache_id=#{project.cache_id}"
       )
--- a/zebra/lib/zebra/workers/job_request_factory.ex
+++ b/zebra/lib/zebra/workers/job_request_factory.ex
@@ -95,6 +95,8 @@ defmodule Zebra.Workers.JobRequestFactory do
          {:ok, open_id_token_env_vars} <-
            OpenIDConnect.load(job, repo_env_vars, organization, project, job_type, spec_env_vars),
          {:ok, callback_token} <- CallbackToken.generate(job) do
+      log_dropped_cache_vars(job, project, cache_env_vars)
+
       org_url = "https://#{organization.org_username}.#{Application.get_env(:zebra, :domain)}"
 
       env_vars =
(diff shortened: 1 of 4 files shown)


zebra/lib/zebra/workers/job_request_factory.ex
```diff
@@ -195,4 +197,13 @@ defmodule Zebra.Workers.JobRequestFactory do
       common_vars
     end
   end
+
+  defp log_dropped_cache_vars(job, project, cache_env_vars) do
+    if cache_env_vars == [] and not Job.self_hosted?(job.machine_type) and
```
dexyk, line 204: @skipi Fires on intentional forked-PR skips. Every forked PR with the flag on logs a false "cache env vars not injected" warning.

zebra/lib/zebra/workers/job_request_factory/cache.ex
```diff
@@ -31,32 +28,58 @@ defmodule Zebra.Workers.JobRequestFactory.Cache do
...
+    cond do
+      response.status.code != ok_code ->
+        Logger.warning(
+          "Cachehub describe returned non-OK status. cache_id=#{cache_id} status_code=#{inspect(response.status.code)}"
+        )
+
+        failed(:non_ok_response)
+        {:ok, nil}
+
+      is_nil(response.cache) or response.cache.credential == " " ->
```
dexyk: @skipi `== " "` misses `""`, was this intentional?

zebra/lib/zebra/workers/job_request_factory.ex
skipi: Fixed in e24665f2. The warning now gates on `Cache.forked_pr?(repo_proxy)`, so intentional forked-PR skips (flag-gated in `Cache.find/3`) no longer log a false "cache env vars not injected". Made `forked_pr?/1` public and added direct unit tests (forked / same-org / no-PR / nil).

zebra/lib/zebra/workers/job_request_factory/cache.ex
skipi: Good catch — not intentional. Replaced `== " "` with a trim-based `blank?/1` that treats `nil`, `""`, and any whitespace as blank, so an empty credential can no longer slip through and write an empty key file. Added an empty-string-credential test. Fixed in e24665f2.

dexyk (approved):

PR #1039, opened 2026-05-26
fix(ee/rbac): issues when readiing saml user

DamjanBecirovic:
## 📝 Description

Fix SAML JIT re-provisioning for users who were previously removed from an organization.

When a SamlJitUser record already existed, login skipped provisioning and never re-checked org membership. This change re-runs JIT provisioning for existing SAML users when they are no longer part of the org, so they are added back correctly on login.

## ✅ Checklist

### DIFF
--- a/ee/rbac/lib/rbac/okta/saml/api.ex
+++ b/ee/rbac/lib/rbac/okta/saml/api.ex
@@ -123,7 +123,8 @@ defmodule Rbac.Okta.Saml.Api do
     {:ok, email, attributes} = PayloadParser.parse(integration, params, consume_uri, metadata_uri)
 
     with {:ok, scim_saml_user} <- find_scim_or_saml_user(integration, email),
-         {:ok, scim_saml_user} <- maybe_provision_existing_saml_jit_user(integration, scim_saml_user),
+         {:ok, scim_saml_user} <-
+           maybe_provision_existing_saml_jit_user(integration, scim_saml_user),
          {:ok, user} <- find_user(scim_saml_user),
          {:ok, user} <- FrontRepo.User.set_remember_timestamp(user) do
       Watchman.increment("saml_login.success")
--- a/ee/rbac/lib/rbac/okta/saml/api.ex
+++ b/ee/rbac/lib/rbac/okta/saml/api.ex
@@ -123,6 +123,7 @@ defmodule Rbac.Okta.Saml.Api do
     {:ok, email, attributes} = PayloadParser.parse(integration, params, consume_uri, metadata_uri)
 
     with {:ok, scim_saml_user} <- find_scim_or_saml_user(integration, email),
+         {:ok, scim_saml_user} <- maybe_provision_existing_saml_jit_user(integration, scim_saml_user),
          {:ok, user} <- find_user(scim_saml_user),
          {:ok, user} <- FrontRepo.User.set_remember_timestamp(user) do
       Watchman.increment("saml_login.success")
@@ -209,6 +210,22 @@ defmodule Rbac.Okta.Saml.Api do
     else
       {:ok, user} -> {:ok, user}
     end
+  end
+
+  defp maybe_provision_existing_saml_jit_user(
+         %{jit_provisioning_enabled: true},
+         %Rbac.Repo.SamlJitUser{} = saml_jit_user
+       ) do
+    if saml_jit_user.user_id == nil or
+         !Rbac.RoleManagement.user_part_of_org?(saml_jit_user.user_id, saml_jit_user.org_id) do
+      Rbac.Okta.Saml.JitProvisioner.AddUser.run(saml_jit_user)
+    else
+      {:ok, saml_jit_user}
+    end
+  end
+
+  defp maybe_provision_existing_saml_jit_user(_integration, scim_saml_user) do
+    {:ok, scim_saml_user}
   end
 
   defp expires_at_unix(integration) do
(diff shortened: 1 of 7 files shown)


skipi:
Rebased this branch and prepared it for merge:

- **Rebased onto `main`**, which dropped a stale pre-#1038 zebra delta this branch was carrying.
- **`mix format`** applied to the two changed SAML files (`lib/rbac/okta/saml/api.ex`, `test/rbac/okta/saml/api_test.exs`) — fixes the `RBAC EE: QA → Lint` failure.
- **Stacked on the CVE fix** semaphoreio/semaphore#1043 (cowboy/cowlib/plug/hackney bumps in `ee/rbac`), which clears the `RBAC EE: Deployment Preconditions → Check dependencies` (mix_audit) failure.

**Merge order:** merge #1043 first, then this branch can be rebased onto `main` so its diff collapses to just the SAML change. Until then the dep-bump commit shows in this PR's diff by design (it's the stack base).

ee/rbac/lib/rbac/okta/saml/api.ex
```diff
@@ -123,6 +123,8 @@ defmodule Rbac.Okta.Saml.Api do
     {:ok, email, attributes} = PayloadParser.parse(integration, params, consume_uri, metadata_uri)
 
     with {:ok, scim_saml_user} <- find_scim_or_saml_user(integration, email),
+         {:ok, scim_saml_user} <-
+           maybe_provision_existing_saml_jit_user(integration, scim_saml_user),
```
dexyk: @skipi this re-provision path ignores current assertion claims.

`attributes` parsed at line 123 are not used here. `AddUser.run` derives role/groups from stored `jit_user.attributes` (`add_user.ex:90`, `add_user.ex:131`), which may be stale. A user downgraded in IdP can be re-added with old elevated role/group mapping.

dexyk (changes requested):

skipi:
@dexyk good catch — fixed in dc7d6c0a.

The re-provision path now refreshes the stored `SamlJitUser.attributes` from the **current** assertion (`attributes` parsed in `handle_auth`) before calling `AddUser.run`, so role/group derivation uses present-day claims instead of stale stored ones. Added `SamlJitUser.update_attributes/2` and threaded the parsed attributes through `maybe_provision_existing_saml_jit_user/3`.

Regression test added: provision a JIT user whose IdP role maps to **Admin**, remove them from the org, then re-login with a **downgraded** assertion → asserts the re-added user gets **Member** (not the stale Admin) and the stored attributes are refreshed. Only the JIT path is touched; SCIM/`OktaUser` falls through the catch-all.

ee/rbac/lib/rbac/okta/saml/api.ex
```diff
@@ -211,6 +213,24 @@ defmodule Rbac.Okta.Saml.Api do
...
 
+  defp maybe_provision_existing_saml_jit_user(
+         %{jit_provisioning_enabled: true},
+         %Rbac.Repo.SamlJitUser{} = saml_jit_user,
+         attributes
+       ) do
+    if saml_jit_user.user_id == nil or
+         !Rbac.RoleManagement.user_part_of_org?(saml_jit_user.user_id, saml_jit_user.org_id) do
+      {:ok, saml_jit_user} = Rbac.Repo.SamlJitUser.update_attributes(saml_jit_user, attributes)
+      Rbac.Okta.Saml.JitProvisioner.AddUser.run(saml_jit_user)
```
dexyk, line 224: @skipi  `assign_groups/2` (add_user.ex:145) has no `user_part_of_group?` guard.
Every re-add login enqueues N new `GroupManagementRequest` rows; browser/proxy
retries pile up linearly while the async processor catches up. Please gate on
existing membership.

`AddUser.run/1`'s moduledoc (add_user.ex:3-11) declares `state == :pending`
and `user_id == nil` as preconditions. This call site violates both; it works only
by coincidence. Either update the moduledoc + add a direct re-entrancy test, or
extract `AddUser.reassign(jit_user)` and call that here.

ee/rbac/test/rbac/okta/saml/api_test.exs
```diff
@@ -66,6 +66,96 @@ defmodule Rbac.Okta.Saml.Api.Test do
...
+
+        {:ok, response} = post("/okta/auth", saml_payload("denis@example.com"))
+
+        assert response.status_code == 302
+        assert Rbac.RoleManagement.user_part_of_org?(saml_jit_user.user_id, @org_id)
+        assert_called_exactly(UserJoinedOrganization.publish(saml_jit_user.user_id, @org_id), 2)
+      end
+    end
+
+    test "valid SAML re-provisions a removed JIT user using current assertion claims, not stale stored ones",
```
dexyk, line 103: @skipi Please add a regression test for two rapid SAML re-add posts and assert bounded GroupManagementRequest rows (no duplicate pending add_user requests).

dexyk (changes requested):

skipi:
@dexyk second round addressed in `084a5890`.

**Group-request idempotency (`assign_groups`, add_user.ex:145):** now gated — a group is skipped when the user is already a member (`UserGroupBinding.member?/2`) or an `:add_user` request is already in-flight (`GroupManagementRequest.open_add_user_request_exists?/2`). So re-add logins and browser/proxy retries no longer pile up `GroupManagementRequest` rows.

**`AddUser.run/1` preconditions:** went with making `run/1` genuinely idempotent (role assign already no-ops when in-org; groups now guarded) and updated the moduledoc to drop the false `state==:pending` / `user_id==nil` assumptions and document re-entrancy. No separate `reassign/1` entry point.

**Tests:**
- `add_user_test`: a second `run/1` does not duplicate group add requests.
- `api_test`: two rapid SAML re-add logins enqueue exactly one `:add_user` request per mapped group (no linear pile-up).

dexyk (approved):

PR #1041, opened 2026-05-26
fix(github_hooks): resolve BranchPayload SHA path with a single compare call

cchristous:
## Summary

When `InternalApi::RepoProxy::BranchPayload#call` receives a fully-resolved 40-char SHA, it made **two** GitHub API calls: `GET /repos/:repo/git/refs/heads/:branch` (via `reference`) followed by `GET /repos/:repo/commits/:sha` (via `commit`). On that path the reference lookup is pure overhead — its only outputs are a branch-existence check and a `ref` echo. When the supplied branch no longer exists upstream (e.g. cron-driven periodic tasks pinned to EOL release branches), GitHub returns 404 and a rate-limit unit is burned per trigger.

Production audit-log analysis on one self-hosted Semaphore deployment counted **6,177 such 404s in 2 days across 862 distinct repos**, dominantly from periodic-scheduler triggers — roughly **5.4% of the production GitHub App's rate-limit budget** over that window.

## Approach

An earlier revision of this PR simply *skipped* the reference lookup on the SHA path. As [@dexyk pointed out in review](https://github.com/semaphoreio/semaphore/pull/1041#issuecomment-4690705282), `BranchPayload` also backs the public Workflows API, and that lookup doubles as a branch-existence guard: skipping it meant a valid SHA + a typo'd/deleted branch would silently build on a phantom branch instead of failing fast. Changing that default behavior for all users is undesirable.

This revision keeps the validation but makes it cheaper. On the SHA path, `reference` + `commit` are replaced by a single `compare(base: sha, head: branch)` call:

- **Validation preserved.** `compare` 404s on a missing branch exactly like the old `git/refs/heads/:branch` lookup, so typo'd/deleted branches still fail fast — *not* a silent phantom-branch build.
- **Commit reused.** The compare response's `base_commit` is the commit object for `sha`, so the separate `commit` call is dropped.
- **Net effect:** one GitHub API call instead of two on the SHA path, removing the gratuitous-call pressure that motivated this PR without weakening validation.

The branch/SHA divergence `status` returned by `compare` is intentionally ignored — the prior contract never verified SHA-on-branch membership, so this change does not introduce stricter checking than existed before.

## Changes

`github_hooks/lib/repo_host/github/client.rb`
- Add `RepoHost::Github::Client#compare(repo, base, head)`, delegating to Octokit and mapping GitHub exceptions (incl. 404 → `RepoHost::RemoteException::NotFound`) like the sibling `reference`/`commit` methods.

`github_hooks/lib/internal_api/repo_proxy/branch_payload.rb`
- Rewrite the SHA fast path to resolve via a single `compare` call, sourcing the head commit from `base_commit`.
- The non-SHA path (partial SHAs, empty strings, tag-style inputs) is preserved verbatim — still `reference()` + `commit_sha/2` + `commit()`.

## Why this is safe

- **Non-SHA path untouched.** Behavior is preserved character-for-character when `sha` is not a full 40-char SHA.
- **Public Workflows API contract preserved.** Missing/typo'd branches still 404 → fail fast, directly addressing the review concern.
- **No stricter validation.** Divergence status is ignored, matching the previous contract which never verified SHA-on-branch membership.
- **No payload shape change**, no public API change, no migration needed.

## Tests

- `RepoHost::Github::Client#compare`: delegation + 404 → `NotFound` translation parity coverage.
- `BranchPayload` SHA path: asserts a single `compare(repo, sha, branch)` call (no `reference`, no `commit`), input `ref` echoed, and a well-formed payload built from `base_commit`.
- The parametric fall-through block (non-SHA → `reference`) is unchanged.

## Known unrelated CI failure

The `Policy::BundlerAudit` (security-toolbox) check currently fails on **`net-imap` 0.5.14** (CVE-2026-47240 / 47241 / 47242, published 2026-06-14). This is **not** introduced by this PR:

- The `net-imap` pin comes from #1054 on `main`; this PR touches only `branch_payload.rb` and `client.rb`.
- `Policy::BundlerAudit` pulls the latest advisory DB at run time, so any `github_hooks` build after the advisory dropped trips it regardless of code — this is the first build to hit it.
- It is repo-wide (`main` carries the same pin) and the fix is a dependency bump to `net-imap >= 0.5.15`, best handled as a dedicated `deps(github_hooks)` PR per the convention in #1054 / #1056.

## Related

Companion PR #1040 fixes a separate but adjacent GitHub rate-limit drain in `repository_hub` (recognising the 422 "max statuses per SHA+context" response shape). The same audit-log analysis surfaced both bugs.


dexyk:
Hi @cchristous, the `github_hooks` workflow has been fixed on the `main` branch. Please rebase your branch onto `main` so that the checks can run correctly.

cchristous:
> Hi @cchristous, the `github_hooks` workflow has been fixed on the `main` branch. Please rebase your branch onto `main` so that the checks can run correctly.

Done - CI is passing now

dexyk:
Thanks @cchristous for the contribution. The changes work as described and reduce the rate of API calls.

One blocker, though: `BranchPayload` also backs the public Workflows API, so skipping validation isn't cron-only, a valid SHA + a nonexistent/typo'd branch now silently builds on a phantom branch instead of failing fast. That hurts all users, so we'd rather not change the default.

Two possible workarounds:
- Put the new behavior behind an env var (off by default) so it can be enabled on self-hosted instances that want it.
- Use `compare(sha, branch)` to verify the SHA actually belongs to the branch, same single API call, but it keeps the validation.

cchristous:
Good catch — agreed that silently building on a phantom branch is the wrong default, and that this path isn't cron-only.

I went with your second suggestion: the SHA path now uses a single `compare(sha, branch)` call. It keeps the branch-existence validation (still 404s on a missing/typo'd branch, so no behavior change there), and reuses the compare response's `base_commit` so we drop the separate `commit` call too — net one API call instead of two, without the original's loss of validation.

The divergence `status` is intentionally ignored, so this doesn't introduce stricter SHA-on-branch checking than the previous contract had. Updated the title and description to match.

github_hooks/lib/repo_host/github/client.rb
```diff
@@ -135,6 +135,12 @@ def commit(repo, sha)
       handle_octokit_exceptions(exception)
     end
 
+    def compare(repo, base, head)
+      user_client.compare(repo, base, head)
```
dexyk: @cchristous `compare` goes through `user_client`, which is built with `auto_paginate: true`. Octokit's `compare` paginates the `commits[]` list, so when the SHA is far behind the branch head this fires many requests to walk commits we never read. That can increase rate-limit usage.

We only need `base_commit` (and optionally `status`/`behind_by`), all first-page envelope fields, pagination isn't required. Proposed fix: route `compare` through a dedicated client with `auto_paginate: false` (a separate memoized client, not by toggling `user_client.auto_paginate`, which would race the shared client).

Same method: `head` is interpolated into the URL unescaped, so branch names with URL-significant chars (e.g. `feat#1`) break, escape it per `/`-segment like the old `reference` path did.

dexyk (changes requested):

github_hooks/lib/repo_host/github/client.rb
cchristous: Both addressed in 0701b4d6:

- **Pagination:** `compare` now routes through a dedicated, memoized `non_paginating_client` (`auto_paginate: false`). `base_commit` is a first-page envelope field, so this resolves in one request regardless of how far the SHA is behind the branch head. Kept it as a separate client rather than toggling `user_client.auto_paginate`, per your race concern.
- **Escaping:** added `escape_ref`, which escapes each `/`-separated segment (so `feat#1` → `feat%231`) while leaving the slashes in `release/1.0` literal.

Added client specs covering the non-paginating routing and the escaping (slashes preserved), plus the existing 404 → `NotFound` mapping.

github_hooks/lib/internal_api/repo_proxy/branch_payload.rb
```diff
@@ -9,11 +9,33 @@ def initialize(ref, sha)
...
+      #     the dedicated `commit` call.
+      #
+      # Net: one GitHub API call instead of two. The branch/SHA divergence
+      # `status` is intentionally ignored — we only use `compare` for branch
+      # existence + the commit object, matching the prior contract (which never
+      # verified SHA-on-branch membership either).
+      if SHA_REGEXP.match?(sha)
+        branch_name = ref.delete_prefix("refs/heads/")
+        response_ref = ref
+        branch_commit = repo_host.compare(repo, sha, branch_name).base_commit
```
dexyk: @cchristous Passing the bare branch name to `compare` reintroduces the missing-branch regression: GitHub resolves bare refs with git precedence (tags before heads), so a deleted branch with a same-named tag validates against the tag and builds.

dexyk (changes requested):

github_hooks/lib/internal_api/repo_proxy/branch_payload.rb
cchristous: Good catch — fixed in 1ad08296. The head is now the fully-qualified `refs/heads/<branch>` (the input `ref`, which `PayloadFactory` guarantees is `^refs/heads/`-prefixed) rather than the bare name, so GitHub resolves it strictly in the heads namespace and 404s on a deleted branch regardless of any same-named tag. Updated the spec to assert the qualified ref is passed.

cchristous: Follow-up: I verified the qualified-ref behaviour empirically against the live GitHub API (public repo, unauthenticated):

- `compare/<sha>...refs/heads/main` → `200` with `base_commit` populated
- `compare/<sha>...refs/heads/<nonexistent>` → `404`

So `refs/heads/<branch>` resolves strictly in the heads namespace and still fails fast on a missing branch — a same-named tag at `refs/tags/<name>` is never picked up, which closes the precedence gap you flagged.

Also hardened the specs (fdcce76a): one pins that the payload's author/email/username/message are wired from `compare`'s `base_commit` identically to the old `commit()` path, and one asserts `BranchPayload#call` propagates a `compare` 404 (`RemoteException::NotFound`) rather than swallowing it or building a phantom payload.

dexyk (approved):

dexyk:
@cchristous, please rebase this onto the latest `main` branch so that the CI workflow can pass. Thanks!

cchristous:
> @cchristous, please rebase this onto the latest `main` branch so that the CI workflow can pass. Thanks!

Done

PR #1042, opened 2026-05-26
fix(github_hooks): make create_ref idempotent, drop probe-then-create in ensure_ref

cchristous:
## Summary

`Semaphore::RepoHost::Hooks::Handler.ensure_ref` uses a probe-then-create pattern: call `repo_host.reference(...)` first, and on `NotFound` fall through to `repo_host.create_ref(...)`. The 404 is load-bearing — it's the signal to create the ref — so every PR-processing call pays for one "speculative" GET that is guaranteed to 404 on the first invocation for a given SHA.

In audit-log analysis of one self-hosted deployment, this is the dominant source of `GET /repos/:repo/git/refs/semaphoreci/<sha>` 404s. Direct analog of PR #893's pattern (replacing probe-then-act with a single authoritative call), applied to the Ruby/Octokit code path instead of the Elixir/tentacat one.

## What this PR does

`github_hooks/lib/repo_host/github/client.rb`:

- New `ref_already_exists?(exception)` predicate that matches `Octokit::UnprocessableEntity` with body containing `"Reference already exists"` — mirrors the existing `hook_exists?` / `maximum_number_of_statuses?` predicates.
- `handle_octokit_exceptions` translates that 422 into a new typed `RepoHost::RemoteException::ReferenceAlreadyExists`, consistent with how the file already handles the other 422 cases.

`github_hooks/lib/semaphore/repo_host/hooks/handler.rb`:

- `ensure_ref` calls `create_ref` directly and rescues `ReferenceAlreadyExists` to treat that case as success.
- Emits `Watchman.increment("github_hooks.ensure_ref.ref_already_exists")` and a `Logman.info` line on the rescue path, so the idempotent case is alertable instead of silent.

## Why this is safe

- **`ensure_ref`'s caller discards the return value.** `update_pr_data` (handler.rb:295-302) constructs its own result tuple using the `ref` *string*, not the API response object. Both the old probe path (returning a Sawyer resource) and the new create path (returning a Sawyer resource or `nil`) satisfy the contract.
- **Error envelope unchanged.** The outer `rescue RepoHost::RemoteException::Unauthorized, RepoHost::RemoteException::NotFound` clause still triggers `:without_reference` for those cases — they're raised by `handle_octokit_exceptions` for the corresponding Octokit exceptions, regardless of which client method triggered them.
- **Only the "Reference already exists" 422 is swallowed at the `ensure_ref` layer.** Any other 422 (e.g. "Invalid object SHA") still raises `RepoHost::RemoteException::Unknown` so callers see the failure — confirmed by a dedicated negative test case. `TooManyRequests` and other transient errors propagate as before.
- **Net rate-limit impact: -1 request per PR processed.** The remaining request is the unavoidable POST `/git/refs` that the function is actually trying to make.

## Tests

- `RepoHost::Github::Client` — four new contexts following the `hook_exists?` pattern:
  - "when create_ref encounters 'Reference already exists' (idempotent)" — raises `ReferenceAlreadyExists`.
  - A second variant exercising the realistic Octokit message format (`POST <url>: 422 - Reference already exists // See: …`) so the regex breaking against the real wire format would be visible.
  - "when create_ref encounters an unrelated 422" — raises `RepoHost::RemoteException::Unknown`.
  - "when create_ref encounters TooManyRequests" — propagates `RepoHost::RemoteException::TooManyRequests` (not swallowed).
- `Semaphore::RepoHost::Hooks::Handler` — the existing "creates ref on github" example is updated to assert `repo_host.reference` is NOT called and pin `create_ref` to `.once`. New examples cover:
  - "treats ReferenceAlreadyExists as success and still proceeds" — `create_ref` raising the typed exception is treated as success and the workflow continues.
  - A shared `without_reference fallback` example that exercises **both** `Unauthorized` and `NotFound` from `create_ref` — asserts `:without_reference` is taken, the "without-reference" log line fires, and the workflow doesn't end up in `STATE_UNAUTHORIZED_REPO` / `STATE_NOT_FOUND_REPO`.
  - "propagates RepoHost::RemoteException::Unknown raised by create_ref" — guards against the rescue accidentally broadening to swallow real failures.

## Test plan

- [x] `rspec` on touched files — all new and modified specs pass.
- [ ] (vendor CI) full github_hooks suite once this PR opens. Note: the `Policy::BundlerAudit` + image-trivy checks flag pre-existing CVEs in `net-imap`, `zlib`, `libgnutls30` etc. — unrelated to this change (they're transitive deps and base-image packages).

## Related

Companion PRs from the same audit-log analysis:
- #1040 — `fix(repository_hub): cache GitHub max-statuses 422 to stop rate-limit drain` (Elixir / tentacat side)
- #1041 — `fix(github_hooks): skip git/refs lookup in BranchPayload when caller supplied a SHA`


dexyk:
Hi @cchristous, the `github_hooks` workflow has been fixed on the `main` branch. Please rebase your branch onto `main` so that the checks can run correctly.

cchristous:
> Hi @cchristous, the `github_hooks` workflow has been fixed on the `main` branch. Please rebase your branch onto `main` so that the checks can run correctly.

Done. CI is passing now

dexyk (approved):

PR #1046, opened 2026-05-29
feat(zebra): gate resilient git checkout behind git_clone_slow_retry feature flag

skipi:
## What

Adds a per-organization feature flag, `git_clone_slow_retry`, that — when enabled — injects `SEMAPHORE_GIT_CLONE_SLOW_RETRY=true` into the job environment built by `JobRequestFactory`.

This is the **producer** side of the toolbox checkout-resiliency work (semaphoreci/toolbox#538 + follow-up semaphoreci/toolbox#539). The toolbox `checkout` reads `SEMAPHORE_GIT_CLONE_SLOW_RETRY` to opt into slow-clone detection + resilient retry (speed monitoring, retries, alternative-endpoint fallback), and is a no-op when the var is absent. So the rollout is: flip the flag for an org → zebra injects the var → toolbox enables the behavior. Default off = unchanged behavior.

## How

Mirrors the existing `TestResults` / `:test_results_no_trim` pattern exactly:

- New `Zebra.Workers.JobRequestFactory.GitCheckout` module with `env_vars/1`, gated on `FeatureProvider.feature_enabled?(:git_clone_slow_retry, param: org_id)`.
- Appended to the env-var list in `JobRequestFactory` alongside `TestResults.env_vars(org_id)`.
- Registered in the test `StubbedProvider` (hidden by default, enabled for a dedicated test org id).

Deliberately *not* threading `org_id` through `Repository.*` — that would churn the exact-match repository tests for no benefit. Only the on/off switch is injected; the toolbox owns the tuning defaults (threshold / timeout / grace / retries).

## Tests

`zebra/test/zebra/workers/job_request_factory/git_checkout_test.exs` — feature disabled → `[]`; enabled → `SEMAPHORE_GIT_CLONE_SLOW_RETRY=true`.

## Follow-up (not in this PR)

The `git_clone_slow_retry` feature must be registered in the feature management backend before it can be toggled per-org in production.

### DIFF
--- a/zebra/lib/zebra/workers/job_request_factory/git_checkout.ex
+++ b/zebra/lib/zebra/workers/job_request_factory/git_checkout.ex
@@ -1,0 +1,30 @@
+defmodule Zebra.Workers.JobRequestFactory.GitCheckout do
+  @moduledoc """
+  Handles resilient git checkout configuration for jobs.
+
+  When the :git_clone_slow_retry feature is enabled for an organization,
+  this module adds the SEMAPHORE_GIT_CLONE_SLOW_RETRY environment variable,
+  which opts the toolbox `checkout` into slow-clone detection and resilient
+  retry (speed monitoring, retries, and alternative-endpoint fallback).
+
+  The toolbox keeps sensible defaults for the tuning knobs
+  (threshold/timeout/grace/retries), so only the on/off switch is injected
+  here; the feature is a no-op in the toolbox when this var is absent.
+  """
+
+  alias Zebra.Workers.JobRequestFactory.JobRequest
+
+  @doc """
+  Returns environment variables for resilient git checkout.
+
+  If the :git_clone_slow_retry feature is enabled for the organization,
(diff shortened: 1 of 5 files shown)


zebra/lib/zebra/workers/job_request_factory/git_checkout.ex
```diff
@@ -0,0 +1,30 @@
...
+
+  @doc """
+  Returns environment variables for resilient git checkout.
+
+  If the :git_clone_slow_retry feature is enabled for the organization,
+  adds SEMAPHORE_GIT_CLONE_SLOW_RETRY=true to enable it.
+  """
+  def env_vars(org_id) do
+    if FeatureProvider.feature_enabled?(:git_clone_slow_retry, param: org_id) do
+      [JobRequest.env_var("SEMAPHORE_GIT_CLONE_SLOW_RETRY", "true")]
```
dexyk, line 32: @skipi I would suggest injecting it only when the job is running on cloud agents, and not on self-hosted ones.

dexyk (approved):

PR #1048, opened 2026-06-01
feat(public-api): record per-client request metrics

skipi:
## What

Add a `ClientMetrics` plug to the v1alpha public API router that emits a Watchman metric (timing + per-status counter) and log fields for every request, tagged by the calling client.

## Why

Requests from different clients (first-party CLIs, the MCP server, generic API consumers) are currently indistinguishable in metrics. This adds a low-overhead per-request metric so traffic can be sliced by client surface, command, and version in dashboards — and header-less callers are tagged `source=api`, so the metric covers all traffic, not only first-party tooling.

## How

Reads the `x-semaphore-client-*` request headers:

- tags `[source, command, version]`; missing headers default to `source=api`, `command=na`, `version=na`
- the statsd_graphite backend keeps only three positional tags, so request **status** is encoded in the metric **name** suffix (`.ok` / `.client_error` / `.server_error`) rather than consuming a tag slot
- `command` and `version` are sanitised against a tight allow-list before becoming tags, to bound metric cardinality
- the plug is registered globally on the router (right after the request logger); `Plug.Logger` is extended to log the same client fields

## Tests

`test/util/client_metrics_test.exs` (7 tests) covering tag building, defaulting, source validation, sanitisation, status-suffixed names, and log rendering. Existing `request_metrics_test.exs` still passes. Compiles clean under `--warnings-as-errors`; `mix format` clean.

### DIFF
--- a/projecthub-rest-api/lib/projecthub/client_metrics.ex
+++ b/projecthub-rest-api/lib/projecthub/client_metrics.ex
@@ -4,7 +4,7 @@ defmodule Projecthub.ClientMetrics do
   ingestion turns these into metrics — no statsd/Watchman involved.
 
   Event keys: severity, message, client_source, client_command, client_version,
-  client_org, status, duration_ms.
+  client_org_id, status, duration_ms.
   """
 
   alias Plug.Conn
@@ -31,7 +31,7 @@ defmodule Projecthub.ClientMetrics do
         client_source: source(conn),
         client_command: command(conn),
         client_version: version(conn),
-        client_org: org(conn),
+        client_org_id: org_id(conn),
         status: conn.status,
         duration_ms: duration
       }
@@ -50,18 +50,14 @@ defmodule Projecthub.ClientMetrics do
   end
 
   @doc """
-  Org identity from the auth-set headers (`x-semaphore-org-username`, else
-  `x-semaphore-org-id`). Returns `"na"` when unauthenticated.
+  Org identity from the auth-set `x-semaphore-org-id` header (trusted, stable).
+  Returns `"na"` when unauthenticated.
   """
-  @spec org(Conn.t()) :: String.t()
-  def org(conn) do
-    name = header(conn, "x-semaphore-org-username")
-    id = header(conn, "x-semaphore-org-id")
-
-    cond do
-      is_binary(name) and name != "" -> name
-      is_binary(id) and id != "" -> id
-      true -> @na
+  @spec org_id(Conn.t()) :: String.t()
+  def org_id(conn) do
+    case header(conn, "x-semaphore-org-id") do
+      id when is_binary(id) and id != "" -> id
+      _ -> @na
     end
   end
 
--- a/projecthub-rest-api/lib/projecthub/client_metrics.ex
+++ b/projecthub-rest-api/lib/projecthub/client_metrics.ex
@@ -25,6 +25,9 @@ defmodule Projecthub.ClientMetrics do
      client_source, client_command, client_version, client_org_id, status,
      duration_ms. Carries the full-fidelity values (raw dotted version) that
      don't belong in metric tags.
+
+  Health-check / ingress-probe paths are skipped so kubelet probes don't
+  inflate the source=api counters or the log volume.
   """
 
   alias Plug.Conn
--- a/projecthub-rest-api/lib/projecthub/client_metrics.ex
+++ b/projecthub-rest-api/lib/projecthub/client_metrics.ex
@@ -1,60 +1,47 @@
 defmodule Projecthub.ClientMetrics do
   @moduledoc """
-  Per-request metric attributing projecthub-rest-api traffic by client (sem-ai
-  CLI/MCP vs generic `api`), command, and version — the same contract as the
-  v1alpha plug. Reads the `x-client-*` headers sem-ai attaches; header-less
-  callers tag `source=api`, so the metric covers all traffic, not only sem-ai.
+  Emits one structured JSON log line per request to stdout. Downstream log
+  ingestion turns these into metrics — no statsd/Watchman involved.
 
-  Emits per request:
-
-    * `Projecthub.router.client_request` (timing) + `.<status>` (counter) — this
-      app's per-service rich metric, tags `[source, command, version]`. Status is
-      in the metric NAME because the statsd_graphite backend keeps only 3
-      positional tags.
-    * `api.client_usage` `[source]` and `api.org_usage` `[org, source]` — generic,
-      service-agnostic counters emitted with the *same* names by every backend,
-      so they aggregate cross-service (the Watchman `service` tag, from this app's
-      prefix, splits per backend). `org` comes from the auth-set
-      `x-semaphore-org-*` headers (trusted, bounded cardinality).
-
-  Tag values are graphite-safe: `.`/`+` -> `_` so a value can't inject the carbon
-  path separator and corrupt the measurement. `command`/`version` are also
-  regex-sanitised to bound cardinality.
+  Event keys: severity, message, client_source, client_command, client_version,
+  client_org, status, duration_ms.
   """
 
   alias Plug.Conn
 
-  @metric "Projecthub.router.client_request"
-  @usage_metric "api.client_usage"
-  @org_metric "api.org_usage"
   @na "na"
   @known_sources ~w(semai-cli semai-mcp)
   @command_regex ~r/\A[a-z0-9_-]{1,50}\z/
   @version_regex ~r/\A[A-Za-z0-9._+-]{1,30}\z/
 
-  @doc "Register a before_send hook that submits the timing + per-status counter."
+  @doc """
+  Register a before_send hook that emits a JSON event line once the response
+  status is known.
+  """
+  @spec track_request(Conn.t()) :: Conn.t()
   def track_request(conn) do
     start = System.monotonic_time(:millisecond)
-    tags = client_tags(conn)
-    [src | _] = tags
 
     Conn.register_before_send(conn, fn conn ->
       duration = System.monotonic_time(:millisecond) - start
-      Watchman.submit({@metric, tags}, duration, :timing)
-      Watchman.increment({metric_name(conn.status), tags})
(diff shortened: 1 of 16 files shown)


public-api/v1alpha/lib/pipelines_api/util/client_metrics.ex
```diff
@@ -0,0 +1,189 @@
...
+
+  def status_label(status) when is_integer(status) and status >= 500, do: "server_error"
+  def status_label(status) when is_integer(status) and status >= 400, do: "client_error"
+  def status_label(status) when is_integer(status), do: "ok"
+  def status_label(_status), do: "unknown"
+
+  defp command(conn), do: sanitize(header(conn, "x-client-command"), @command_regex)
+  defp version(conn), do: sanitize(header(conn, "x-client-version"), @version_regex)
+
+  defp sanitize(value, regex) when is_binary(value) do
```
dexyk, line 145: @skipi command/version are validated for charset+length only, not for an enumerated set, then used as Graphite tag values. The public API is free/self-service, `x-client-*` is not stripped by auth, and there's no dedup, so any caller can mint unbounded distinct series on the shared metrics backend.

projecthub-rest-api/lib/projecthub/client_metrics.ex
```diff
@@ -0,0 +1,111 @@
...
+
+    cond do
+      is_binary(name) and name != "" -> graphite_safe(name)
+      is_binary(id) and id != "" -> graphite_safe(id)
+      true -> nil
+    end
+  end
+
+  defp command(conn), do: sanitize(header(conn, "x-client-command"), @command_regex)
+  defp version(conn), do: sanitize(header(conn, "x-client-version"), @version_regex)
```
dexyk, line 110: @skipi command/version are validated for charset+length only, not for an enumerated set, then used as Graphite tag values. The public API is free/self-service, `x-client-*` is not stripped by auth, and there's no dedup, so any caller can mint unbounded distinct series on the shared metrics backend.

public-api-gateway/api/middleware/client_metrics_middleware.go
```diff
@@ -0,0 +1,111 @@
...
+	return "api"
+}
+
+func clientCommand(r *http.Request) string {
+	return sanitizeTag(r.Header.Get("x-client-command"), commandRegex)
+}
+
+func clientVersion(r *http.Request) string {
+	return sanitizeTag(r.Header.Get("x-client-version"), versionRegex)
+}
```
dexyk, line 112: @skipi command/version are validated for charset+length only, not for an enumerated set, then used as Graphite tag values. The public API is free/self-service, `x-client-*` is not stripped by auth, and there's no dedup, so any caller can mint unbounded distinct series on the shared metrics backend.

public-api/v1alpha/lib/pipelines_api/util/client_metrics.ex
```diff
@@ -0,0 +1,189 @@
...
+  (`x-semaphore-org-username` preferred for readability, else `x-semaphore-org-id`).
+  These are set by the auth service (trusted), not the client, so cardinality is
+  bounded by real orgs. Returns nil when unauthenticated (no org present).
+  """
+  def org_tag(conn) do
+    name = header(conn, "x-semaphore-org-username")
+    id = header(conn, "x-semaphore-org-id")
+
+    cond do
+      is_binary(name) and name != "" -> graphite_safe(name)
```
dexyk: @skipi `org_tag` prefers the mutable `x-semaphore-org-username` over the stable `x-semaphore-org-id`. An org rename splits one customer across two series and corrupts long-term per-org dashboards.

projecthub-rest-api/lib/projecthub/client_metrics.ex
```diff
@@ -0,0 +1,111 @@
...
+  Org identity from the auth-set headers (`x-semaphore-org-username`, else
+  `-org-id`). Auth-set (trusted), not client, so cardinality is bounded by real
+  orgs. nil when unauthenticated.
+  """
+  def org_tag(conn) do
+    name = header(conn, "x-semaphore-org-username")
+    id = header(conn, "x-semaphore-org-id")
+
+    cond do
+      is_binary(name) and name != "" -> graphite_safe(name)
```
dexyk: @skipi `org_tag` prefers the mutable `x-semaphore-org-username` over the stable `x-semaphore-org-id`. An org rename splits one customer across two series and corrupts long-term per-org dashboards.

public-api-gateway/api/middleware/client_metrics_middleware.go
```diff
@@ -0,0 +1,111 @@
...
+	if v != "" && re.MatchString(v) {
+		return graphiteSafe(v)
+	}
+	return naTag
+}
+
+// orgTag uses the auth-set headers (trusted, never client-supplied);
+// username preferred for readability, org id as fallback. "" when unauthenticated.
+func orgTag(r *http.Request) string {
+	if name := r.Header.Get("x-semaphore-org-username"); name != "" {
```
dexyk: @skipi `org_tag` prefers the mutable `x-semaphore-org-username` over the stable `x-semaphore-org-id`. An org rename splits one customer across two series and corrupts long-term per-org dashboards.

dexyk (changes requested):

public-api/v1alpha/lib/pipelines_api/util/client_metrics.ex
skipi: I think this is fine - otherwise we would have only org_ids in the panels which make it harder to identify who's the biggest api consumer.

public-api-gateway/api/middleware/client_metrics_middleware.go
```diff
@@ -0,0 +1,111 @@
...
+		return func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
+			start := time.Now()
+			rec := NewResponseRecorder(w)
+
+			next(rec, r, pathParams)
+
+			src := clientSource(r)
+			tags := []string{src, clientCommand(r), clientVersion(r)}
+
+			_ = watchman.BenchmarkWithTags(start, clientRequestMetric, tags)
```
dexyk: @skipi `BenchmarkWithTags` sends microseconds, but the Elixir plugs emit milliseconds. Same `client_request` timer ends up 1000x off between the gateway and the Elixir services.

skipi:
Reworked this — client attribution is now off statsd metrics entirely (c4fccfde). Each request emits a single structured JSON `client_request` log event (`client_source` / `client_command` / `client_version` / `client_org` / `status` / `duration_ms`, plus `client_trace` on v1alpha); metrics are derived downstream from the logs.

That addresses both threads:

- **Cardinality** — `command`/`version` are no longer metric tags, they're log fields. Nothing unbounded or client-supplied goes into a metric series anymore, so there's no way to mint series on the shared backend. Bounding now happens downstream (a log-based metric with a filtered, enumerated command set) where it's a single editable place.
- **Timer units** — the gateway no longer calls `BenchmarkWithTags` (go-watchman is removed); duration is just `duration_ms` in the event, identical across the Elixir services and the gateway.
- **org_tag** — it's a log field now, not a tag, so the username-vs-id point no longer splits any series. Kept username for readability per the thread.

PTAL.

public-api/v1alpha/lib/pipelines_api/util/client_metrics.ex
skipi: Reconsidered this — switched to the stable `client_org_id` (from the trusted `x-semaphore-org-id`) and dropped the username, so org renames no longer split the series. Readable names are resolved at display time. Done in 3a809d5c.

projecthub-rest-api/lib/projecthub/client_metrics.ex
skipi: Reconsidered this — switched to the stable `client_org_id` (from the trusted `x-semaphore-org-id`) and dropped the username, so org renames no longer split the series. Readable names are resolved at display time. Done in 3a809d5c.

public-api-gateway/api/middleware/client_metrics_middleware.go
skipi: Reconsidered this — switched to the stable `client_org_id` (from the trusted `x-semaphore-org-id`) and dropped the username, so org renames no longer split the series. Readable names are resolved at display time. Done in 3a809d5c.

public-api/v1alpha/lib/pipelines_api/util/client_metrics.ex
```diff
@@ -0,0 +1,91 @@
...
+        client_source: source(conn),
+        client_command: command(conn),
+        client_version: version(conn),
+        client_org_id: org_id(conn),
+        client_trace: trace_id(conn) || @na,
+        status: conn.status,
+        duration_ms: duration
+      }
+
+      IO.puts(Jason.encode!(event))
```
dexyk: @skipi please check if `Jason` is included as a runtime dependency here. It seems to be present only in dev/test through `credo` / `ecto`.

dexyk (changes requested):

skipi:
Update: `4dc02c25` re-introduces the Watchman metrics on all three surfaces, emitted alongside the structured `client_request` log event (which stays). Deriving metrics from logs downstream turned out to be a poor fit — log-based metrics are charged by label cardinality — so we're emitting directly again, and we monitor cardinality on the metrics backend so we can iterate quickly if a tag misbehaves.

Differences from the metric code this PR previously removed:

- the gateway timer is submitted in **milliseconds** via `TimingWithTags` — `BenchmarkWithTags` emits microseconds, which was the 1000x unit mismatch against the Elixir plugs that @dexyk caught
- `configureMetrics` stays backgrounded, so the gateway never blocks boot waiting on statsd
- the org tag uses the auth-set stable org id (same value the log event carries), not the mutable username

Tag values remain regex-sanitised (charset + length) and graphite-safe (`.`/`+` → `_`).

dexyk (dismissed):

dexyk (approved):

PR #1063, opened 2026-06-15
feat(ppl): optimize init-job checkout (blobless + sparse) behind a feature flag

DamjanBecirovic:
## What

The pipeline initialization (compilation) job only needs the pipeline YAML and
the Git history (commits/trees, used by `change_in`) to compile a pipeline — not
the full repository working tree. For large repositories the full checkout
dominates the init-job runtime (cloning hundreds of MiB and materializing tens
of thousands of files on every run).

This makes the init job perform a **blobless partial clone + sparse checkout** of
the pipeline directory, gated by a per-organization feature flag and disabled
when pre-flight checks are present.

## How it works

The init job emits `SEMAPHORE_GIT_PARTIAL_CLONE_FILTER=blob:none` and
`SEMAPHORE_GIT_SPARSE_CHECKOUT_PATHS=<working_dir>` before `checkout`. The
optimization is applied only when **both** hold:

- there are **no pre-flight checks** (their custom commands may rely on the full
  working tree), and
- the `sparse_checkout_init_job` feature is enabled for the organization.

The feature check **fails closed** — a missing org id or an unreachable Feature
service keeps the standard full checkout. `change_in` keeps working because it
relies only on `git diff --name-only` / `--shortstat` / `merge-base`, which need
tree/commit objects (present in a blobless clone), not blob contents.

## Changes

- `proto`: generate `InternalApi.Feature` stubs; add `INTERNAL_API_LOCAL_PATH`
  support to the proto Makefile so they can be regenerated from a local
  `internal_api` checkout.
- `ppl`: add the `feature_provider` dependency and a FeatureHub-backed provider
  + gRPC client (`INTERNAL_API_URL_FEATURE`); `Ppl.Features` exposes the
  fail-closed flag check; init `FeatureProvider` and a `:feature_cache` in the
  application supervisor.
- `ppl`: gate the optimized checkout in the compilation init-job command
  generation; auto-disable when pre-flight checks are configured.
- Build: move ppl's Docker build context to the repository root (matching the
  top-level services) so `feature_provider`, which lives outside `plumber/`, is
  included in the image.
- Tests: feature-flag gate + provider coverage, with a Feature gRPC mock; plus a
  small fix to make an unrelated integration test's env restore nil-safe.

> Note: `yaml_elixir` is pinned to `~> 1.3` via override because the umbrella's
> YAML validators rely on 1.x return semantics; this means feature_provider's
> YAML provider is not used here (the FeatureHub gRPC provider is). A follow-up
> is tracked to align on yaml_elixir 2.x for YAML-defined features.

## Dependencies

This relies on the toolbox `checkout` partial/sparse support and the spc
`commands_file` on-demand fetch being released (and the toolbox image rebuilt).

## Testing

- `Ppl: QA` and `Ppl: Integration QA` green; dev and prod image builds validated.

## Related PRs

- https://github.com/semaphoreci/toolbox/pull/542
- https://github.com/semaphoreci/spc/pull/59


plumber/ppl/config/config.exs
```diff
@@ -31,6 +31,14 @@ config :vmstats,
   sink: Ppl.VmstatsWatchmanSink,
   interval: 10_000
 
+# Feature flags (FeatureHub). The Feature service address is read at call time
+# from INTERNAL_API_URL_FEATURE (see Ppl.FeatureClient). Results are cached in
+# the :feature_cache Cachex instance started by Ppl.Application.
+config :ppl,
+  feature_provider:
```
dexyk, line 38: @DamjanBecirovic we also need to support the YAML feature provider here

dexyk (changes requested):

dexyk (approved):

PR #1075, opened 2026-06-25
fix(zebra): isolate self-hosted dispatch per organization

skipi:
## What

The dispatcher (`DbWorker` polling jobs in the `scheduled` state) drains queued jobs per machine-type key. Cloud jobs are isolated per `{machine_type, machine_os_image}` — each gets its own per-tick batch, so one type cannot starve another. Self-hosted jobs had that isolation **off**: they were drained as a single global FIFO across **all organizations**, ordered by `scheduled_at`. One organization with a large backlog therefore head-of-line-blocked every other organization's self-hosted jobs (a backlog of a couple thousand queued agents could starve others' self-hosted jobs for minutes).

This change isolates the self-hosted path per `{organization_id, machine_type, machine_os_image}`. Each key gets its own `records_per_tick` batch per tick, so a big backlog in one organization no longer delays others.

## Changes

- **`query_machine_types/1` (`:self_hosted`)** now selects `DISTINCT {organization_id, machine_type, machine_os_image}` (was `{machine_type, machine_os_image}`).
- **New `query_jobs/2` clause** matching the 3-element self-hosted key, adding `where organization_id == ^org_id` alongside the existing `machine_type` / `machine_os_image` filters + `order_by scheduled_at` + `limit`.
- **New `submit_batch_size/3` clause** for the 3-element key (the existing 2-tuple clause would not match a 3-tuple and would raise). The metric tag shape is unchanged (`"<machine_type>-<os_image>"`, no org → no cardinality change).
- **`isolate_machine_types`** is now always `true` for the dispatcher. It was already true for cloud and the dev `:all` mode; only the self-hosted-only deployment had it forced off. The previous env-conditional helper is removed.

## Cloud path is unchanged

The `:cloud` branch of `query_machine_types/1`, the `nil` and 2-tuple `query_jobs/2` clauses, the `FOR UPDATE SKIP LOCKED` claim, and all job state transitions are untouched. A regression test locks in that cloud dispatch stays org-blind (isolated only by `{machine_type, machine_os_image}`).

## Supporting index (bundled)

Add a partial composite index, created `CONCURRENTLY` (jobs is a large, hot, shared table):

```
on jobs (machine_type, organization_id, machine_os_image, scheduled_at)
WHERE aasm_state = 'scheduled'
```

Rationale: without it, each per-key drain query scans the entire scheduled set (via the partial `aasm_state` index), filters `machine_type` + `organization_id` in memory, then sorts — i.e. `O(#keys × scheduled-set)` per tick once isolation fans out. This index makes each per-key query a direct range scan (no in-memory filter, no sort) and also serves the per-tick `DISTINCT`.

## Per-tick fan-out

Distinct key counts are modest (tens), and with the index each per-key query is a cheap range scan, so the existing serial `Enum.each` over keys is kept — no cap, no extra parallelism added. The `dispatcher.tick.duration` Watchman benchmark already wraps the whole tick and can be watched post-deploy.

## Deploy

No manifest change. Self-hosted isolation is gated by the existing `DISPATCH_SELF_HOSTED_ONLY` deployment; this is a normal image roll.

## Tests

Extended the dispatcher test harness:
- self-hosted jobs are isolated per `{org, type, os_image}` — one org's older/larger backlog does **not** block another org's jobs of the same type name (this test fails on the pre-change global-FIFO behaviour and passes after).
- cloud dispatch is **not** org-isolated (unchanged behaviour) — regression guard.

Results (run via `make test.ex`, `mix test --warnings-as-errors`, Postgres 9.6):
- `test/zebra/workers/dispatcher_test.exs`: **14 tests, 0 failures**
- full zebra suite: **497 tests, 0 failures**

## Unrelated smell noticed (not fixed here)

`Dispatcher.duration_bucket/1` maps `sec >= 3 and sec < 10` to `"from_0s_to_3s"` — looks like a copy/paste slip (should be a `3s..10s` bucket). Left untouched to keep this PR focused; flagging for a separate fix.

### DIFF
--- a/zebra/priv/legacy_repo/migrations/20260625090000_add_scheduled_dispatch_index_at_jobs_table.exs
+++ b/zebra/priv/legacy_repo/migrations/20260625090000_add_scheduled_dispatch_index_at_jobs_table.exs
@@ -1,0 +1,14 @@
+defmodule Zebra.LegacyRepo.Migrations.AddScheduledDispatchIndexAtJobsTable do
+  use Ecto.Migration
+  @disable_migration_lock true
+  @disable_ddl_transaction true
+
+  def change do
+    create index(:jobs, [:machine_type, :organization_id, :machine_os_image, :scheduled_at],
+      name: "index_jobs_on_machine_type_org_image_when_scheduled",
+      concurrently: true,
+      where: "aasm_state = 'scheduled'",
+      if_not_exists: true
+    )
+  end
+end
--- a/zebra/priv/legacy_repo/migrations/20260625090000_add_scheduled_dispatch_index_at_jobs_table.exs
+++ b/zebra/priv/legacy_repo/migrations/20260625090000_add_scheduled_dispatch_index_at_jobs_table.exs
@@ -4,11 +4,12 @@ defmodule Zebra.LegacyRepo.Migrations.AddScheduledDispatchIndexAtJobsTable do
   @disable_ddl_transaction true
 
   def change do
-    create index(:jobs, [:machine_type, :organization_id, :machine_os_image, :scheduled_at],
-      name: "index_jobs_on_machine_type_org_image_when_scheduled",
-      concurrently: true,
-      where: "aasm_state = 'scheduled'",
-      if_not_exists: true
+    create_if_not_exists(
+      index(:jobs, [:machine_type, :organization_id, :machine_os_image, :scheduled_at],
+        name: "index_jobs_on_machine_type_org_image_when_scheduled",
+        concurrently: true,
+        where: "aasm_state = 'scheduled'"
+      )
     )
   end
 end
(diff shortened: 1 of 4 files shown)


zebra/priv/legacy_repo/migrations/20260625090000_add_scheduled_dispatch_index_at_jobs_table.exs
```diff
@@ -0,0 +1,14 @@
...
+  @disable_ddl_transaction true
+
+  def change do
+    create index(:jobs, [:machine_type, :organization_id, :machine_os_image, :scheduled_at],
+      name: "index_jobs_on_machine_type_org_image_when_scheduled",
+      concurrently: true,
+      where: "aasm_state = 'scheduled'",
+      if_not_exists: true
+    )
+  end
```
loadez, line 14: This will need manual deployment (around 15~30 min)

loadez (dismissed):

zebra/priv/legacy_repo/migrations/20260625090000_add_scheduled_dispatch_index_at_jobs_table.exs
skipi: Agreed — the plan is to run this migration manually (CREATE INDEX CONCURRENTLY on jobs, ~15–30 min) *before* the deploy, so the rollout itself picks up an already-built index. Rebased on main; note main now also carries `index_jobs_on_machine_type_scheduled_at_scheduled` (semaphoreio/semaphore#1115) — different columns/name, both coexist.

loadez (approved):

PR #1098, opened 2026-07-08
Fix(Rbac, Front, guard, github_hooks) limit github account to be linked to only one user account

adbatista:
## 📝 Description

**Problem**
 - Multiple Semaphore users could connect to the same GitHub account — nothing enforced uniqueness on `repo_host_accounts`, causing nondeterministic user lookups by provider UID, broken login flows, and duplicate users created during OIDC signup.

 **Changes**
 - Add an application-level uniqueness check in `guard` (mirrored in `ee/rbac`) that rejects connecting a GitHub account already linked to another user, covering the OAuth connect callback and user-creation paths.
 - Surface a clear "account already connected to another Semaphore user" error on the account page via a new `account_taken` OAuth error code.
 - OIDC signup skips the conflicting provider link (with a warning logged) instead of failing user creation; gRPC `User.Create` fails with `invalid_argument` and rolls back.

## ✅ Checklist


guard/lib/guard/front_repo/federated_identity_sync_request.ex
```diff
@@ -0,0 +1,131 @@
...
+      set: [
+        attempts: attempts,
+        last_error: String.slice(error, 0, @max_error_length),
+        next_attempt_at: retry_at,
+        updated_at: now()
+      ]
+    )
+
+    :ok
+  end
```
loadez, line 97: @adbatista Is it ok for us to retry this forever? Also is lease_due able to starve this if we get x failed rows?

ee/rbac/lib/rbac/front_repo/repo_host_account.ex
```diff
@@ -255,10 +395,15 @@ defmodule Rbac.FrontRepo.RepoHostAccount do
...
+    unrevoke? = unrevoke_transition?(changeset)
+    result = FrontRepo.update(changeset)
 
     case result do
       {:ok, account} ->
+        account = if unrevoke?, do: release_revoked_uid_rows(account), else: account
+
         Logger.info(
           "Successfully updated RepoHostAccount for #{account.user_id} from #{inspect(account)} to #{inspect(data)}"
         )
```
loadez, line 411: @adbatista can you check what kind of data is this logging? We may want to remove or clean this. There are multiple log lines like this in this file

github_hooks/lib/internal_api/repository_integrator/repository_integrator_server.rb
```diff
@@ -174,6 +174,27 @@ def update_revoke_status(rha)
...
+        end
+      end
+
+      def uid_actively_held_by_other?(rha)
+        ::RepoHostAccount
+          .where(:repo_host => rha.repo_host, :github_uid => rha.github_uid)
+          .where.not(:id => rha.id)
+          .where(:revoked => false)
+          .exists?
+      end
```
loadez, line 217: We need a grace period like we have in guard here (guard/lib/guard/front_repo/repo_host_account.ex)

PR #1099, opened 2026-07-09
Address new decimal and websocket-driver dependency advisories

skipi:
Two advisories landed in today's advisory-DB refresh and fail `Check dependencies` on any branch whose changes wake the affected blocks:

- **websocket-driver 0.8.0** (github_hooks, transitive): CVE-2026-54463 / CVE-2026-54464 — patched upstream, bumped to 0.8.1 (existing `>= 0.6.1` constraint already allows it).
- **decimal 2.1.1** (plumber/spec, transitive via jason): GHSA-rhv4-8758-jx7v — patched only in decimal 3.0, which jason pins out. Added `plumber/spec/.mix-audit.txt` accepting it with the same rationale already ratified in `plumber/ppl/.mix-audit.txt` (decimal unused in app code). The spec app never had this file because its CI block hadn't run until a recent branch touched it.

### DIFF
--- a/github_hooks/Gemfile.lock
+++ b/github_hooks/Gemfile.lock
@@ -487,7 +487,7 @@ GEM
       hashdiff (>= 0.4.0, < 2.0.0)
     webrick (1.9.2)
     websocket (1.2.11)
-    websocket-driver (0.8.0)
+    websocket-driver (0.8.1)
       base64
       websocket-extensions (>= 0.1.0)
     websocket-extensions (0.1.5)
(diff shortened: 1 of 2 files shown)


skipi:
Superseded by semaphoreio/semaphore#1101 — same websocket-driver bump; for decimal it removes the unused lock entry outright instead of accepting the advisory via .mix-audit.txt (decimal is only an optional transitive of jason, never fetched). Note: the ignore-file comment here references ../ppl/.mix-audit.txt, which this PR does not add.

PR #1100, opened 2026-07-09
fix(deps): resolve dependency audit failures (websocket-driver, decimal)

skipi:
## What

Fixes the `Check dependencies` (`check.ex.deps` / `check.ruby.deps`) failures currently blocking PRs that touch `plumber/spec` or `github_hooks`.

- **github_hooks**: bump `websocket-driver` 0.8.0 → 0.8.2 (transitive, via actioncable). Fixes CVE-2026-54463, CVE-2026-54464, CVE-2026-54465 and GHSA-2x63-gw47-w4mm — memory-exhaustion / DoS vectors in the protocol parser. Lockfile-only change (`bundle lock --update=websocket-driver`).
- **plumber/spec**: remove the stale `decimal` 2.1.1 entry from `mix.lock`, flagged by GHSA-rhv4-8758-jx7v (unbounded exponent DoS, patched only in 3.0.0). Nothing in the app requires decimal — it is only an optional dependency of jason — so the leftover lock entry is dropped instead of bumped. `mix deps.get --check-locked` passes without it.

## Verification

- `bundler-audit check --update` in `github_hooks`: **No vulnerabilities found** (advisory DB 2026-07-08)
- `mix_audit` in `plumber/spec` (elixir:1.16 container): **No vulnerabilities found**, and `mix deps.get --check-locked` resolves cleanly

### DIFF
--- a/github_hooks/Gemfile.lock
+++ b/github_hooks/Gemfile.lock
@@ -487,7 +487,7 @@ GEM
       hashdiff (>= 0.4.0, < 2.0.0)
     webrick (1.9.2)
     websocket (1.2.11)
-    websocket-driver (0.8.0)
+    websocket-driver (0.8.2)
       base64
       websocket-extensions (>= 0.1.0)
     websocket-extensions (0.1.5)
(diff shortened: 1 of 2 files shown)


skipi:
Duplicate of semaphoreio/semaphore#1101 (byte-identical diff).

PR #1112, opened 2026-07-15
feat(guard): extend MCP OAuth token TTL to 24h

adbatista:
## 📝 Description

Extends the MCP OAuth access token TTL from 1 hour to 24 hours in guard, the OAuth authorization server for mcp.<domain>. 

The token endpoint now passes an explicit ttl_seconds to JWT.create_token/2 and returns the same value in expires_in, keeping the JWT exp claim and the response field in lockstep so clients schedule re-authentication correctly. 

Since the flow issues no refresh tokens, this reduces full browser re-auth for MCP clients from hourly to at most once a day — a stopgap until refresh-token support lands separately.

## ✅ Checklist

### DIFF
--- a/guard/lib/guard/mcp_oauth/token.ex
+++ b/guard/lib/guard/mcp_oauth/token.ex
@@ -11,8 +11,6 @@ defmodule Guard.McpOAuth.Token do
   alias Guard.Store.McpOAuthAuthCode
   alias Guard.McpOAuth.{JWT, PKCE}
 
-  @access_token_ttl_seconds 86_400
-
   @doc """
   Exchanges an authorization code for an access token.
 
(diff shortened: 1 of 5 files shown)


guard/lib/guard/mcp_oauth/token.ex
```diff
@@ -11,6 +11,8 @@ defmodule Guard.McpOAuth.Token do
   alias Guard.Store.McpOAuthAuthCode
   alias Guard.McpOAuth.{JWT, PKCE}
 
+  @access_token_ttl_seconds 86_400
```
skipi: Thoughts on moving this value to the config file? It will be easier to manage in the future.

skipi (changes requested):

guard/lib/guard/mcp_oauth/token.ex
adbatista: 🤔 yeah, makes more sense, having it in a config file

skipi (dismissed):

skipi (approved):

PR #1113, opened 2026-07-16
fix(github_notifier): enforce commit status ordering per check

adbatista:
## 📝 Description
GitHub commit statuses for fast pipelines could arrive out of order (success before pending), permanently wedging the check on "pending", because the running/done AMQP consumers built statuses independently and POSTed them via fire-and-forget tasks with no ordering guarantee. 

This PR routes every status for the same check (repository/sha/pipeline/context) through the same worker in a small supervised pool — serial, in-order delivery — and drops a stale pending once a terminal status is already sent. Failed sends are no longer cached as delivered; they now raise an exception, so the message is redelivered and retried instead of being silently lost.

**Crash safety:** durability lives in RabbitMQ, not in the sender processes. A message is acked only after handle_message returns, and delivery is now synchronous end-to-end (GenServer.call blocks until the status is actually sent), so if a worker crashes or the pod dies mid-delivery, the message remains unacked and is redelivered — notify re-runs against the live pipeline state and resends. Redeliveries are idempotent (dedupe cache + terminal guard), so the worst case is a duplicate identical status on GitHub, not a lost one. The old fire-and-forget code acked before the POST ran, so a crash silently dropped statuses; this changes delivery to at-least-once.

## ✅ Checklist


github_notifier/lib/github_notifier/application.ex
```diff
@@ -16,26 +16,24 @@ defmodule GithubNotifier.Application do
...
+      ] ++
+        filter_enabled([
+          {{GRPC.Server.Supervisor, {grpc_services(), 50_051}}, enabled?("START_API")},
+          {{Services.BlockFinishedNotifier, []}, enabled?("START_CONSUMERS")},
+          {{Services.PipelineStartedNotifier, []}, enabled?("START_CONSUMERS")},
+          {{Services.PipelineFinishedNotifier, []}, enabled?("START_CONSUMERS")},
+          {{Services.PipelineSummaryAvailableNotifier, []}, enabled?("START_CONSUMERS")},
+          {{GithubNotifier.FeatureProviderInvalidatorWorker, []}, true},
+          feature_provider()
+        ])
```
adbatista: Previously, the consumers started before the caches, so a message arriving during boot could crash on a not-yet-started dependency (:noproc) and only recover via the 10s delay-queue retry. Starting the infrastructure first guarantees every dependency exists before the first message is consumed, and gives the reverse on shutdown: consumers stop first, dependencies last.

skipi:
@codex review

loadez (commented):
Solid approach overall. The Postgres FOR UPDATE lease + fencing token + per-pod GenServer serialization is the right shape for cross-instance ordering, and the race coverage is good. A few things before this merges, one blocking.

**Blocking**

1. Fail-open on transient DB errors. When the guard query hits any transient error against the front DB (not only the migration window), the code falls through and sends anyway. The guard table lives in the same front DB that's already the saturation point, so a front-DB blip makes every github_notifier sender lose ordering at once and reproduces exactly the out-of-order / stuck pending status this PR is meant to fix. That branch also has no test. Either fail-closed (hold + retry) on transient errors, or make the fail-open path explicit and bounded, and cover it with a test.

**Non-blocking**

2. `:busy` drops the channel and reconnects, which thrashes under load. Prefer a bounded retry/backoff over dropping the channel.
3. The 90s lease is long relative to the retry budget; a stuck holder eats most of the window before anyone else proceeds. Consider a shorter lease + renewal.
4. Supervisor strategy change (one_for_one -> rest_for_one) widens the blast radius of a single child crash. Confirm that's intended.
5. Stale generated Go proto stub in the diff; regenerate or drop it.

Also needs a rebase, it's conflicting against main.

loadez:
The fail-open and channel-drop fixes look good, that closes the main concerns cleanly and the redelivery-forcing test is exactly the right shape.

One follow-up worth a look before this lands: in `server.ex` the `finalize/2` return value is discarded (around line 182). `finalize/2` shares the same `transaction/1` rescue as `claim/1`, so it can return `{:error, reason}` on a transient DB error, but nothing checks it. Sequence: delivery succeeds, `finalize` hits a DB blip, `last_state` never gets written and `claimed_at` never clears, yet the caller still returns `code: :OK` and the message acks. The row then sits with a stale/nil `last_state` and a live lease; once the lease expires, a redelivered stale PENDING for that check no longer matches `last_state in @terminal_states` and gets delivered, reproducing a narrower version of the pending-after-terminal wedge this PR is closing. Tighter window than the original fail-open (needs the blip to land on `finalize` plus a stale redelivery after expiry), so not a merge blocker, but worth at least logging the `finalize` error, and maybe refreshing the lease rather than silently dropping the bookkeeping. Same pattern on `release/2`'s return in the rescue branch, lower stakes (only costs up to ~90s of extra lease hold, no ordering impact).

loadez (approved):

PR #1125, opened 2026-07-28
Fix(github_notifier) Set Alpine version to 3.23

adbatista:
## ✅ Checklist


k-patrik:
There is no need to downgrade OTP. Instead pin Alpine to 3.23 (musl 1.2.5) since musl 1.2.6 in Alpine 3.24 rejects the BEAM's 8Kib.
This is of course temporary until fixed in upstream. Complete details in https://github.com/erlang/otp/pull/11376

adbatista:
Thanks @k-patrik I'll check this

dexyk (approved):

PR #1141, opened 2026-08-05
Store role ids in deployment target ROLE subject rules

### LINKED ISSUE #1140: Deployment target ROLE subject rules created via the public API grant no access (name stored where Gofer expects a role id)
## Summary

A `ROLE` subject rule created or updated through the public API (v1alpha **or** v2) silently grants nobody access. The API stores the role's **name** in `subject_id`, while Gofer authorizes by comparing `subject_id` against role **IDs**. The write succeeds, the rule is visible in `GET`, and every promotion attempt is then denied.

This is silent: there is no error at write time and no indication on the target that the rule is inert. A target that previously allowed a role can be left allowing nobody by an unrelated `PATCH`, because `subject_rules` is revalidated and rewritten on every update.

## Root cause

`public-api/v1alpha/lib/pipelines_api/deployments_client/request_formatter.ex`:

```elixir
defp role_name_to_role_map(project_scope_roles) do
  Enum.into(project_scope_roles, %{}, fn role ->
    {String.downcase(role.name), role.name}   # <-- value should be role.id
  end)
end
```

That map is what `normalize_roles/2` substitutes into `subject_id`, so the stored value is the role name.

`ee/gofer/lib/gofer/deployment/guardian.ex` treats it as an ID:

```elixir
defp role_rules_apply?(rules, subject, opts) do
  role_ids = Enum.map(rules, & &1.subject_id)

  case RBAC.check_roles(subject, role_ids, opts) do
    {:ok, assignments} -> Enum.any?(assignments, &elem(&1, 1))
    {:error, _reason} -> false
  end
end
```

`public-api/v2/lib/internal_clients/deployment_targets_client/request_formatter.ex` has the same defect. The intent is visible there — the call site names the map `role_name_to_uuid` and binds `role_id`:

```elixir
defp role_rule!(role_name, role_name_to_uuid) do
  case Map.get(role_name_to_uuid, String.downcase(role_name)) do
    nil -> raise ArgumentError, "Role #{role_name} does not exist in the project."
    role_id -> %API.SubjectRule{type: :ROLE, subject_id: role_id}
  end
end
```

— but it is handed `role.name`, so `role_id` is a name.

## Second, related problem: the API is not round-trippable

Rules created in the web UI store a real role UUID. Writing one back is rejected, because validation only accepts role *names*:

```
PATCH /api/v1alpha/deployment_targets/{id}
{"subject_rules":[{"type":"ROLE","subject_id":"8af7d76c-5687-4ab7-acef-33f99af9f2b2"}]}

400  "role \"8af7d76c-5687-4ab7-acef-33f99af9f2b2\" is not valid"
```

Since `subject_rules` is replaced wholesale, a client cannot do read-modify-write on any target that has a UI-created role rule — the only way to make the `PATCH` succeed is to downgrade the role rule to a name, which silently destroys the access it granted. This is how we hit it: adding a bookmark parameter to 29 targets removed Admin deploy access from all of them, and nothing in any response indicated it.

## Reproduction

```bash
# 1. Write a role rule by name — accepted
curl -X PATCH -H "Authorization: Token $TOKEN" -H "Content-Type: application/json" \
  --data '{"subject_rules":[{"type":"ROLE","subject_id":"Admin"}],"unique_token":"'$(uuidgen)'"}' \
  "https://<org>.semaphoreci.com/api/v1alpha/deployment_targets/$TARGET_ID"

# 2. Read it back — stored as the name, not the id
curl -s -H "Authorization: Token $TOKEN" \
  --get --data-urlencode "project_id=$PROJECT_ID" --data-urlencode "target_name=$NAME" \
  "https://<org>.semaphoreci.com/api/v1alpha/deployment_targets"
# => {"type": "ROLE", "subject_id": "Admin"}

# 3. A user holding the project Admin role attempts the promotion
# => "You don't have rights to deploy to <target> Deployment Target"
```

Supplying `role_id` alongside the name does not help (the name is still what is persisted); `role_id` alone returns 500.

Validation resolves against project-scope roles only, which confirms the lookup is real but discards the id: `Contributor` and `Participant` are accepted, `Owner` and `Member` are rejected as `role "owner" is not valid`.

## Impact

Any `ROLE` subject rule written through either public API version grants no access. Worse, because `PATCH` revalidates and rewrites the whole rule list, any API write to a target — even one that only touches `bookmark_parameter1` — converts a working UI-created role rule into an inert one.

## Fix

PR to follow: store `role.id` and accept a role id as well as a name, in both API versions.

Note that this does not migrate rules already stored as names; those need rewriting once the fix is in.

sid2700:
Fixes #1140.

## The bug

A `ROLE` subject rule written through the public API stored the role's **name** in `subject_id`:

```elixir
defp role_name_to_role_map(project_scope_roles) do
  Enum.into(project_scope_roles, %{}, fn role ->
    {String.downcase(role.name), role.name}   # value should be role.id
  end)
end
```

Gofer authorizes by comparing `subject_id` against role **ids**:

```elixir
# ee/gofer/lib/gofer/deployment/guardian.ex
defp role_rules_apply?(rules, subject, opts) do
  role_ids = Enum.map(rules, & &1.subject_id)
  case RBAC.check_roles(subject, role_ids, opts) do
```

So the rule matched nobody. The write returned 200, `GET` reported the rule, and every promotion was denied — with no error and nothing on the target to show it was inert.

Present in both v1alpha and v2. The v2 call site already expected an id — it names the map `role_name_to_uuid` and binds `role_id` before building `%API.SubjectRule{type: :ROLE, subject_id: role_id}` — it was simply handed `role.name`.

## The change

`role_name_to_role_map/1` now maps to `role.id`, and since the same map validates the incoming `subject_id`, it also accepts a role id:

```elixir
defp role_name_to_role_map(project_scope_roles) do
  Enum.reduce(project_scope_roles, %{}, fn role, acc ->
    acc
    |> Map.put(String.downcase(role.name), role.id)
    |> Map.put(String.downcase(role.id), role.id)
  end)
end
```

Accepting the id matters for more than symmetry. Rules created in the web UI store a real role uuid; sending one back was rejected as `role "<uuid>" is not valid`. Because `subject_rules` is replaced wholesale on update, a client could not do read-modify-write on such a target at all — the only way to make the request succeed was to downgrade the role rule to a name, which silently destroyed the access it granted. That is how we found this: adding a bookmark parameter to 29 production targets removed Admin deploy access from every one of them, and no response indicated anything had changed.

Role names continue to work exactly as before.

## Tests

Updated the assertions that encoded the old behaviour — they expected `subject_id: "Admin"` — to expect the Admin project role's id, looked up from the stub DB. Added a case asserting a role rule given as an id round-trips unchanged.

- `public-api/v1alpha/test/deployments_client_test.exs`
- `public-api/v1alpha/test/router/deployments/create_test.exs`
- `public-api/v1alpha/test/router/deployments/update_test.exs`
- `public-api/v2/test/router/deployment_targets/create_test.exs`

I don't have an Elixir toolchain set up, so **I have not run the suite locally** — I'd appreciate CI confirming it, and I'm happy to iterate on anything it flags.

## Notes for reviewers

- **No data migration.** Rules already stored as names stay inert until rewritten. A migration in Gofer that resolves stored role names to ids would fix existing installations; I left it out as it seemed a separate call. Happy to add it.
- **v2 response shape.** v2 exposes `spec.subject_rules.roles` as the raw `subject_id` list, so those now read as ids rather than names — the same thing v1alpha already returns for UI-created rules. If you would rather v2 kept presenting names, the tidier variant is to resolve id → name on read and I can do that instead.
- Validation resolves against project-scope roles only. That is unchanged here, and worth knowing: `Contributor` and `Participant` are accepted while `Owner` and `Member` are rejected, which is what confirmed the lookup was real but discarding the id.

### DIFF
--- a/ee/gofer/lib/gofer/deployment/model/deployment/subject_rule.ex
+++ b/ee/gofer/lib/gofer/deployment/model/deployment/subject_rule.ex
@@ -32,8 +32,16 @@ defmodule Gofer.Deployment.Model.Deployment.SubjectRule do
   defp validate_subject_id(:USER, changeset = %Ecto.Changeset{}),
     do: Ecto.Changeset.validate_required(changeset, [:subject_id])
 
-  defp validate_subject_id(:ROLE, changeset = %Ecto.Changeset{}),
-    do: Ecto.Changeset.validate_required(changeset, [:subject_id])
+  defp validate_subject_id(:ROLE, changeset = %Ecto.Changeset{}) do
+    changeset
+    |> Ecto.Changeset.validate_required([:subject_id])
+    |> Ecto.Changeset.validate_change(:subject_id, fn :subject_id, value ->
+      case Ecto.UUID.cast(value) do
+        {:ok, _} -> []
+        :error -> [subject_id: "must be a role id"]
+      end
+    end)
+  end
 
   defp validate_subject_id(:GROUP, changeset = %Ecto.Changeset{}),
     do: Ecto.Changeset.validate_required(changeset, [:subject_id])
--- a/ee/gofer/test/deployment/model/deployment_test.exs
+++ b/ee/gofer/test/deployment/model/deployment_test.exs
@@ -43,7 +43,7 @@ defmodule Gofer.Deployment.Model.DeploymentTest do
     end
 
     test "when name is longer than 255 characters then invalid", %{bare_params: params} do
-      name = for _ <- 1..300, into: "", do: <<Enum.random('0123456789abcdef')>>
+      name = for _ <- 1..300, into: "", do: <<Enum.random(~c"0123456789abcdef")>>
       changeset = assert_invalid?(&Deployment.changeset/2, %{params | name: name})
 
       assert [
@@ -107,6 +107,12 @@ defmodule Gofer.Deployment.Model.DeploymentTest do
 
     test "casts subject rules", %{bare_params: params, subject_rules: rules} do
       assert_valid?(&Deployment.changeset/2, %{params | subject_rules: rules})
+    end
+
+    test "when ROLE subject rule carries a role name instead of an id then invalid",
+         %{bare_params: params} do
+      rules = [%{type: :ROLE, subject_id: "Admin"}]
+      assert_invalid?(&Deployment.changeset/2, %{params | subject_rules: rules})
     end
 
     test "casts object rules", %{bare_params: params, object_rules: rules} do
--- a/public-api/v1alpha/lib/pipelines_api/deployments_client/request_formatter.ex
+++ b/public-api/v1alpha/lib/pipelines_api/deployments_client/request_formatter.ex
@@ -639,8 +639,10 @@ defmodule PipelinesAPI.DeploymentTargetsClient.RequestFormatter do
   defp check_role_subject_ids(error, _org_id, _subject_rules), do: error
 
   defp role_name_to_role_map(project_scope_roles) do
-    Enum.into(project_scope_roles, %{}, fn role ->
-      {String.downcase(role.name), role.name}
+    Enum.reduce(project_scope_roles, %{}, fn role, acc ->
+      acc
+      |> Map.put(String.downcase(role.name), role.id)
+      |> Map.put(String.downcase(role.id), role.id)
     end)
   end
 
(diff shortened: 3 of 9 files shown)


skipi:
@sid2700 Thank you for the report and the PR 🙇🏻‍♂️ . I've added additional validations and made a documentation change of api v2. Once the blocking PR https://github.com/semaphoreio/semaphore/pull/1135 is merged, I'll rebase this and merge it as well.

skipi (approved):

sid2700:
Hi @skipi, thanks for the approval. 
I am working on some automation for our usecase which is when I stumbled upon this bug. I really hope this can be merged and deployed soon.

loadez (approved):

PR #1142, opened 2026-08-06
perf(rbac): scope role inheritance CTE to org in has_role

skipi:
## What

`Rbac.RoleManagement.has_role/2` builds a recursive role-inheritance CTE seeded from the **entire** `rbac_roles` table — every organization's roles — before joining against a single subject's role bindings. Each call materializes and sorts the global inheritance tree (multi-MB sorts that spill to temp files at default `work_mem`).

This change parametrizes `Queries.role_inheritance_and_mappings_query/1` with an optional `org_id` and passes the org from `has_role`, so the CTE seeds only that organization's roles (a handful of rows instead of the whole table).

## Why

`SubjectsHaveRoles` batches (e.g. deployment-target subject-rule checks, up to N roles per call, evaluated per pipeline event) issue one `has_role` query per role. Under load this saturates the database: each query pays the full-table CTE cost even though the final predicate only ever inspects one org's roles.

## Scope

- `has_role/2` always has a non-nil org id (input verified before the query) — now scoped.
- `compute_permissions` keeps the unscoped default (`org_id \\ nil`) — behavior unchanged.

## Correctness

Role inheritance and org→project role mappings are defined between roles of the same organization, so seeding with the org's roles yields the same reachable set for that org's bindings as the global seed.

### DIFF
--- a/ee/rbac/.mix-audit.txt
+++ b/ee/rbac/.mix-audit.txt
@@ -39,3 +39,9 @@ GHSA-pj7v-xfvx-wmjq
 # CRLF / header injection via unvalidated `domain` and `path` options (no user-controlled domain/path
 # fed to hackney from lib/)
 GHSA-mp55-p8c9-rfw2
+
+# gun 2.4.1 — HTTP request/response splitting (shared cowboy/gun advisory; the <2.16.0 range is
+# cowboy's numbering and over-matches gun). gun is only pulled in as the grpc client transport and
+# ee/rbac's gRPC traffic is internal service-to-service with no user-controlled request lines or
+# headers. Real fix tracked for the fleet-wide dep-bump branch.
+GHSA-w4f7-4cxr-rv3c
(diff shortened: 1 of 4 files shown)


skipi:
Review follow-up, addressed in 8d2886cf:

- **Cross-org binding invariant**: `verify_input_for_new_role` now fetches the role scoped to the caller's org (`get_role_by_id(role_id, org_id)`), which guards both `assign_role` (write path can no longer persist a cross-org binding) and `has_role` (read path matches the invariant). Two-org regression tests added for both paths. Existing cross-org rows (if any) should be audited separately before/after merge — `SELECT count(*) FROM subject_role_bindings srb JOIN rbac_roles r ON r.id = srb.role_id WHERE r.org_id != srb.org_id`.
- **`rbac_roles(org_id)` index**: agreed, deliberately deferred to a follow-up migration PR — the scoped seed already removes the global CTE expansion/sort, which live measurements show is the dominant cost (~3-5s → ~150-260ms per call under saturated load).
- **Deps gate**: the failing check was unrelated to this diff — the `RBAC EE: Deployment Preconditions` block only runs on `change_in('/ee/rbac')`, so advisories published since the last ee/rbac change (Jul 16) surface here first. Accepted GHSA-w4f7-4cxr-rv3c (gun 2.4.1, shared cowboy/gun advisory whose range over-matches gun) with rationale in `.mix-audit.txt`.

loadez (approved):

PR #1150, opened 2026-08-10
fix(github_notifier): report stopped/canceled pipelines as error, not failure

cchristous:
## What

`GithubNotifier.Utils.State` maps every non-`PASSED`, `DONE` pipeline/block to
the GitHub `failure` commit-status state ("The build failed on Semaphore 2.0."),
including ones whose result is `STOPPED` or `CANCELED`.

This routes `STOPPED`/`CANCELED` (state `DONE`) results to the existing
`:STOPPED` build status instead: `State.extract` / `extract_with_summary` emit a
`"stopped"` token, `Status.map_status/1` translates it to
`InternalApi.Repository.CreateBuildStatusRequest.Status` `:STOPPED`, and
`repository_hub` already surfaces `:STOPPED` to GitHub as the `error`
commit-status state (with the "The build was canceled on Semaphore 2.0."
description). `PASSED`/`FAILED` are unchanged.

## Why

A canceled or stopped pipeline never reached a verdict, so reporting it as a
build `failure` is inaccurate — it wasn't a failed build, it was interrupted.
GitHub's `error` commit-status state (which `:STOPPED` already maps to) reflects
that far better, and — unlike simply skipping the notification — it avoids
leaving a stale `pending` status behind.

The delivery path for `:STOPPED` already exists end-to-end: the
`CreateBuildStatusRequest.Status` proto has a `STOPPED` value, and the GitHub,
GitLab, and Bitbucket adapters in `repository_hub` all already accept `:STOPPED`
and translate it for their provider. The notifier was the one layer that never
emitted it, so this is a small, self-contained change on the producing side.

## Scope

Single change in `github_notifier` (`state.ex` + `status.ex`); no proto or
`repository_hub` changes — those already support `:STOPPED`.

The diff is larger than the behavioral change itself: adding the
`STOPPED`/`CANCELED` branch pushed `extract/1` and `extract_with_summary/2` past
Credo's cyclomatic-complexity limit (9). A separate, behavior-preserving commit
therefore flattens the nested `result`/`state` cases into a `%{state: :DONE}`
function-head guard plus small `verdict/1` and `verdict/2` clauses, bringing the
complexity back under the limit. That commit changes structure only — the
`{state, message}` outputs are identical for every input.

## Testing

Added `state_test.exs` cases covering `STOPPED`/`CANCELED` for both `extract/1`
and `extract_with_summary/2` (including the `extract_with_summary` short-circuit
that ignores the test summary for a canceled build), plus a not-yet-`DONE`
(`STOPPING`) case that must stay `pending`. Verified the module compiles and
exercised `extract` / `extract_with_summary` against these cases. (Full
`mix test` relies on CI — deps are not fetched in the local offline
environment.)

### DIFF
--- a/github_notifier/lib/github_notifier/status.ex
+++ b/github_notifier/lib/github_notifier/status.ex
@@ -79,4 +79,5 @@ defmodule GithubNotifier.Status do
   defp map_status("success"), do: :SUCCESS
   defp map_status("pending"), do: :PENDING
   defp map_status("failure"), do: :FAILURE
+  defp map_status("stopped"), do: :STOPPED
 end
--- a/github_notifier/lib/github_notifier/utils/state.ex
+++ b/github_notifier/lib/github_notifier/utils/state.ex
@@ -11,68 +11,45 @@ defmodule GithubNotifier.Utils.State do
   @message_build_canceled "The build was canceled on Semaphore 2.0."
   @message_build_pending "The build is pending on Semaphore 2.0."
 
-  def extract(pipeline) do
-    case pipeline.result do
-      :PASSED ->
-        case pipeline.state do
-          :DONE -> {@success, @message_build_passed}
-          _ -> {@pending, @message_build_pending}
-        end
+  def extract(%{state: :DONE} = pipeline), do: verdict(pipeline.result)
+  def extract(_pipeline), do: {@pending, @message_build_pending}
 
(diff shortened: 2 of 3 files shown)


adbatista (approved):
thanks

DamjanBecirovic (approved):

PR #1151, opened 2026-08-11
feat(guard): issue MCP OAuth refresh tokens with rotation

MathiasWP:
## 📝 Description

**What's wrong.** MCP OAuth access tokens are minted with a 24h TTL (`MCP_OAUTH_ACCESS_TOKEN_TTL_SECONDS`), but `/mcp/oauth/token` only accepted `grant_type=authorization_code` and the token response carried no `refresh_token` — `grant_types_supported` advertised `authorization_code` alone. There is no silent-renewal path in the protocol, so every MCP client is forced back through the browser sign-in flow once a day.

Peer remote MCP servers (Linear, Sentry, Notion, Atlassian) all advertise `refresh_token`, and several issue *shorter*-lived access tokens than Semaphore does. Short access-token TTL is the correct behaviour — the missing refresh grant is the bug. The MCP authorization spec expects both: short-lived access tokens *and* a refresh token in the token response.

**How it's fixed.** Guard now issues an opaque refresh token alongside every access token and accepts `grant_type=refresh_token`.

- Refresh tokens are stored as sha256 hashes in a new `mcp_oauth_refresh_tokens` table, bound to the issuing `client_id`; plaintext is returned once, at issue time.
- MCP clients are public clients (PKCE, no secret), so OAuth 2.1 rotation rules apply: every refresh consumes its token and issues a successor in the same family, and replaying a consumed token revokes the whole family.
- Rotation locks the row with `SELECT FOR UPDATE` inside the transaction, mirroring the existing single-use auth-code path, so concurrent refreshes serialise across pods.
- `refresh_token` is advertised in the authorization server metadata and in the dynamic client registration response, so clients discover it.
- Expired refresh tokens are swept by the existing MCP OAuth cleanup job. Used rows are kept until they expire — that is what makes replay detection possible.
- TTL is configurable via `MCP_OAUTH_REFRESH_TOKEN_TTL_SECONDS`, default 30 days. Access-token TTL is unchanged.

### Notes for reviewers

- **Strict replay handling.** A client that loses a rotated token (e.g. the token response is lost in transit) will re-present the consumed one, which revokes the family and costs one browser re-auth. The spec makes rotation a MUST for public clients, so this is implemented strictly rather than with a reuse grace window — happy to add leeway if you'd prefer.
- **Refresh does not re-check account status.** It mints from `user_id` exactly as the authorization code exchange does. Hard-deleted users cascade away via the FK, but a deactivated account can keep renewing for up to the refresh TTL. Worth a follow-up if MCP access should be cut sooner than the access-token TTL.
- **Migration** `20260811120000_create_mcp_oauth_refresh_tokens` must be applied with the release.

### Tests

- 8 new end-to-end cases on the token endpoint: rotation, repeat refresh, replay revoking the family, client binding, expired / unknown / missing-parameter errors.
- New store test file for issue / lock / rotate / revoke-family / cleanup.
- Existing metadata, DCR and token-exchange assertions updated for the new grant and response field.

I do not have an Elixir toolchain available where this was written, so the suite has not been run locally — CI is the first real execution. Formatting and Credo rules were checked by inspection.

### Docs

`mcp_server/OAUTH_FLOW.md` gains a refresh sequence diagram, token-model and security-guarantee entries, and a troubleshooting line for the daily re-prompt symptom. The user-facing MCP page notes that OAuth clients renew access in the background.

## ✅ Checklist


MathiasWP:
This is an attempt to fix that i have to re-authenticate the Semaphore MCP every 24 hours, which is annoying. I'm using ToolHive, which makes it a bit easier to manage, but it still stands out as the only MCP tool that i have to frequently re-authenticate on...

loadez (commented):
Thanks for this, really nice first contribution. The rotation logic is solid and I could follow it end to end. I went through it carefully and the important parts all hold up: reuse detection, family revocation, and the rotation atomicity are correct, including under concurrent refreshes (the FOR UPDATE path does the right thing).

Two things worth looking at before it goes in, neither a blocker:

1. Absolute family lifetime. Right now the 30-day expiry is a sliding window that resets on every successful refresh, so a token family an attacker manages to steal and keep rotating never actually expires. Reuse detection catches the real user's next attempt, but if the attacker is the only one refreshing, nothing ages the family out. Might be worth a hard ceiling on top of the sliding window (an absolute max age from first issuance) so a compromised family dies regardless of activity.

2. A concurrency test. Would be great to pin the race you're already guarding: two refreshes of the same token at the same time, one wins and the other is rejected with the family revoked. The setup for this already exists, so it's not much work: `use Guard.RepoCase, async: false` gives you a shared Postgres sandbox with real concurrent connections, and there's a working example to copy in `test/guard/cli_auth_test.exs` ("the attempt cap is strict under concurrent entries of the same code") that fires a batch of `Task.async` calls at a `SELECT FOR UPDATE` transaction and asserts exactly one wins. Same shape as what you'd need here.

Again, great work, this is a clean and useful addition. Happy to help if any of it is unclear.

MathiasWP:
@loadez i applied your changes, thoughts?

PR #1165, opened 2026-08-13
chore(hooks-processor): bump tackle to v0.4.0

loadez:


hooks_processor/Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG ELIXIR_VERSION=1.18.1
-ARG OTP_VERSION=26.2.5.6
-ARG ALPINE_VERSION=3.21.6
+ARG OTP_VERSION=26.2.5.3
```
adbatista, line 2: Why roll back to .3 ?

loadez: Rolled back to `26.2.5.3` because `26.2.5.6` breaks the cold image build. On `.6`, `mix local.hex` fails fetching from `builds.hex.pm` with a TLS `unsupported_certificate` / `key_usage_mismatch` alert (OTP's cert-verification change rejects the hex endpoint's cert). Two cold builds on this branch reproduced it; pinning back to `.3` builds green, and the only Dockerfile diff between the failing and passing commits is the `OTP_VERSION` arg. Proper `.6` fix is a follow-up, not this PR.

adbatista (approved):

skipi (approved):

PR #1167, opened 2026-08-13
chore(hooks-receiver): bump tackle to v0.4.0

loadez:


hooks_receiver/.mix-audit.txt
```diff
@@ -0,0 +1,93 @@
...
+GHSA-grp7-v8xh-rj7h
+
+# --- grpc: authorization bypass via path binding override in HTTP transcoding ---
+# This service's gRPC clients do not implement or enable HTTP transcoding (no
+# google.api.http annotations processed, no transcode plug/gateway configured
+# anywhere in the codebase).
+GHSA-mwr4-5g34-j5cq
+
+# --- gun: HTTP/1.1-only issues (unsolicited 101 upgrade hijack, unbounded response
+# buffer) and HTTP/2 PUSH_PROMISE cross-origin cookie injection ---
```
serviceunit0123-pixel: All Gmail account and their password

skipi:
> **Corrected 2026-08-28.** An earlier revision of this comment claimed seven of the suppressed advisory IDs did not exist. **That was wrong and I apologise.** I tested them against GitHub's *global* advisory database, where they 404. They are real, published *repository-level* advisories on `elixir-plug/plug`, `elixir-mint/mint` and `elixir-mint/hpax`, which propagate to the global DB separately. Five of them describe live, unfixed vulnerabilities in this lockfile. The section is rewritten below and the conclusion is reversed: on that point the file was **more** thorough than my review. The rest of the findings are unchanged.

Reviewed this in depth. The code changes are sound and I have no objection to them — the tackle bump, the `import Plug.Test` fix, the cowboy/plug/ranch bumps and the OTP pin all hold up under checking. My remaining concerns are with `hooks_receiver/.mix-audit.txt`. It is a security-audit artifact that whoever reads it next will trust, and a few of its statements are not accurate for this service.

Of the 31 suppressed IDs: **19 match today**, **7 are already superseded** by this PR's own version bumps, and **5 are real, live vulnerabilities that mix_audit cannot see yet** because its mirror lags the upstream repos. No package in this lock has an unsuppressed unfixed match, so the gate would not be red for any other reason.

## Should come before merge

### 1. The hackney rationale is not true for this service

> `# hackney is only reached transitively via the Sentry client, which posts events to`
> `# the operator-configured SENTRY_DSN endpoint`

hooks_receiver has no Sentry dependency — not in `mix.exs`, not in `mix.lock`, not in `lib/` or `config/`. The only occurrence of the string `sentry` in the entire app is that comment.

hackney 1.20.1 arrives solely via `{:httpoison, "~> 2.0", only: :test}` (`mix.exs:34`) — the only `HTTPoison.*` call sites are in `test/router_test.exs`. `Dockerfile:30` runs `mix deps.get --only $MIX_ENV` at `MIX_ENV=prod`, so hackney never enters the production image.

The six lines are still needed (mix_audit reads `mix.lock` verbatim and has no concept of dep environments), but the truthful justification is strictly stronger than the written one.

### 2. Five suppressions are hiding live vulnerabilities that should be fixed, not suppressed

This is the one I'd most want addressed, and it is the opposite of what I first reported. These five are real and this lockfile is genuinely affected:

| Advisory | Severity | Affected | This PR locks |
|---|---|---|---|
| `GHSA-jj2p-32j7-whj2` (hpax) | **High** | `< 1.0.4` | **1.0.2** |
| `GHSA-c59h-fq4p-r36r` (mint) | **High** | `>= 0.5.0`, fixed `1.9.1` | **1.6.2** |
| `GHSA-qrfr-wh4c-3qhw` (mint) | **High** | `< 1.9.2` | **1.6.2** |
| `GHSA-8pf6-g464-h6h9` (mint) | Medium | `< 1.9.2` | **1.6.2** |
| `GHSA-x3x7-96vm-6h2w` (mint) | Medium | `< 1.9.3` | **1.6.2** |

They do not fire today only because `mirego/elixir-security-advisories` — the DB `security-toolbox/policies/dependencies/mix_audit.rb:39` pulls — has not mirrored them yet. When it does, they will. Listing them ahead of the mirror was the right call; the concern is that suppressing them buries three High-severity issues that a version bump would actually fix (see below).

Both are reachable-only-through-mint, and the file's own argument is that mint is dead code — so the risk today is low. But "dead code" and "fixed" are different claims, and the file should not be the reason nobody bumps these.

One correction to the hpax block's reasoning, independent of the above: it says hpax is "only exercised via gun's HTTP/2 transport". gun does not use hpax at all — its only declared requirement is cowlib, and its HPACK comes from `cow_hpack`. **hpax is pulled in by mint**, the very dependency the file argues two blocks later is dead code. The conclusion holds; the dependency graph in the comment does not.

### 3. Seven entries are already fixed by this PR's own bumps

| Entry | Affected | This PR locks |
|---|---|---|
| `GHSA-jfc2-q6qh-g5x8` | cowboy `>= 2.0.0, < 2.15.0` | **2.18.0** |
| `GHSA-468c-vq7p-gh64` | plug `>= 1.18.0, < 1.18.2` | **1.18.5** |
| `GHSA-95qv-c9g9-rm63` | plug `>= 1.18.0, < 1.18.5` | **1.18.5** |
| `GHSA-wpmj-jh88-rpgm` | plug `>= 1.18.0, < 1.18.5` | **1.18.5** |
| `GHSA-g2wm-735q-3f56` | cowlib `>= 2.9.0, <= 2.16.1` | **2.19.0** |
| `GHSA-hv23-4qp7-8c8r` | cowlib `>= 2.6.0, < 2.16.1` | **2.19.0** |
| `GHSA-84f2-rp86-235p` | cowlib `>= 0.1.0, < 2.16.1` | **2.19.0** |

Worth noting the plug bump landed on exactly 1.18.5, which is the patched version for both of the newer plug advisories — so if that was deliberate, it worked. These seven can be dropped, and their comments justify them as unreachable rather than as fixed, which reads like it was carried over from `front/.mix-audit.txt` where front's older cowboy/cowlib/plug genuinely are affected. (The SPDY one is doubly moot — cowlib 2.16.1 deletes `cow_spdy.erl` outright.)

### 4. The `GHSA-w4f7-4cxr-rv3c` rationale does not apply at gun 2.1.0

> `# The gun-attributed instance is additionally a known false positive: mix_audit`
> `# collapses the advisory's per-package version ranges and matches any gun version`
> `# against cowboy's bound`

The collapse is real — the advisory is filed only under `packages/gun/` with ranges `['< 2.4.0', '< 2.16.0']`, and since there is no gun 2.16.0 the second bound is cowboy's. But that is not why it fires here. **This lock has gun 2.1.0, inside gun's own `< 2.4.0` range.** It would fire with or without the collapse. The false-positive framing was carried over from `github_notifier`, whose gun is 2.4.1, where it genuinely does apply.

Before anyone tries the obvious fix: bumping gun does **not** clear this one. The ranges are OR'd, so every shippable gun version — 2.5.0 included — still matches the collapsed `< 2.16.0` bound.

The suppression is still correct, but on **unreachability, not version**. The defective encoder `cow_http_struct_hd:escape_string/2` is byte-identical in cowlib 2.13.0 through 2.19.0 — it is present in this build, just never invoked. gun 2.1.0 references neither `cow_http_struct_hd` nor `wt_protocol`; cowboy 2.18.0 touches only the decode direction; and this service has no `put_resp_header` calls. That is the argument the comment should make.

### 5. Two cross-references point at things that do not exist

- **`secrethub/.mix-audit.txt` does not exist**, on this branch or on `main`. secrethub is a real Elixir app that runs `make check.ex.deps`, but it has no audit file — there are 22 in the repo and it is not among them.
- **`front/.mix-audit.txt` does not contain `GHSA-w4f7-4cxr-rv3c`.** Only `github_notifier` does (`:29-39`). front also carries only 4 of the 6 hackney IDs cited.

The `GHSA-rv48-qqj5-crxg` → github_notifier citation and the `GHSA-r53j-fjj5-mv77` "accepted fleet-wide" claim both check out.

### 6. The grpc block picks the one argument in it that is false

> `# No compatible fixed version exists upstream`

The section header — "no upstream fix released in any 0.x version (0.3.1-0.11.5)" — is accurate. The body is not: all four grpc advisories list `first_patched_versions: 1.0.0`, and grpc 1.0.0 through 1.0.4 are published on hex.

There was a bulletproof argument available and the block does not use it: `GHSA-6ccx` (gzip decompression bomb), `GHSA-q8gf` (`read_full_body/3` accumulation) and `GHSA-grp7` (`GRPC.Codec.Erlpack.decode/2` calling `:erlang.binary_to_term/1` without `:safe`) are all **server-side** — they need an inbound gRPC request, and this service runs no gRPC server in the release. That disposes of all three cleanly. If a version rationale is wanted alongside it, the real blocker is ours, not upstream's: `grpc_mock` (branch `grpc08`) declares `{:grpc, "~> 0.3"}`, which excludes 1.x.

## A suggestion that would shrink this a lot

Three overrides would remove twelve suppressions by fixing rather than excusing — including all three High-severity items from §2:

```elixir
{:gun, "~> 2.4", override: true},
{:mint, "~> 1.9", override: true},
{:hpax, "~> 1.0.4", override: true},
```

- **mint 1.9.3+** clears all eight mint advisories (`x3x7` needs 1.9.3, `8pf6`/`qrfr` 1.9.2, `c59h` 1.9.1).
- **hpax 1.0.4** clears the High-severity HPACK issue.
- **gun 2.4.1** clears `GHSA-2j82-37xg-f9wp`, `GHSA-36w4-95hv-5vwg` and `GHSA-r53j-fjj5-mv77`, all `< 2.4.0`. (`GHSA-w4f7` survives per §4, but at gun 2.4.1 it becomes the same genuine false positive `github_notifier` documents — which would finally make that citation true.) gun 2.4.1 is already proven in-fleet.

The gun bump also resolves a latent inconsistency worth fixing on its own: gun 2.1.0's lock entry pins `{:cowlib, "2.13.0"}` **exactly**, while `{:cowlib, "~> 2.19", override: true}` forces it onto cowlib 2.19.0 — six minors past what gun declares. gun relaxed that pin to `>= 2.15.0` only in 2.2.0. It is safe today (no cowlib export gun uses was removed between 2.13.0 and 2.19.0; the diff is additive, and only the exported type `cow_http:version/0` moved, a dialyzer warning at most) — but it is safe by luck, and moving gun makes it safe by construction.

Between the 7 superseded entries and these bumps, the file would go from 31 entries to roughly 12, every one genuinely firing and carrying an accurate justification.

## Smaller things

- "runs no gRPC server" is true for the release, but `test/test_helper.exs:4` starts one on port 50051 under `MIX_ENV=test`. A parenthetical would stop a future reader thinking the claim is contradicted.
- The title says v0.4.0; the branch actually lands **v0.4.1**.
- Fragility for whoever edits this file next: mix_audit does not trim retained lines and requires `#` at column 0, so an indented comment, an inline trailing comment, or trailing whitespace silently becomes a non-matching "ID". The current file is clean on all three.

For what it's worth, the reachability analysis is good and I checked it rather than assuming: multipart (`router.ex:12` really is `parsers: [:urlencoded, :json]`), no `put_resp_header`/`put_resp_cookie` anywhere, no SSE, no SPDY, JSON-not-protobuf on the inbound path, no `:adapter` passed to any `GRPC.Stub.connect` (so gun, not mint), gun forcing `protocols: [:http2]`, and transcoding being server-only in grpc 0.9.0. It is the citations and the copied-over rationales that need the pass.

skipi (approved):

skipi (approved):

PR #1175, opened 2026-08-13
chore(projecthub): bump tackle to v0.4.1

loadez:


adbatista (dismissed):

loadez:
Re-requesting review: the branch was force-pushed so the prior approval was dismissed. This addresses the Codex P1 on the hackney suppression. The rationale was corrected (hackney is reached via Tentacat in user.ex and deploy_key.ex, not only via Sentry); hackney is bumped 1.18.0 to 1.25.0, which fixes GHSA-9fm9-hp7p-53mf and GHSA-vq52-99r9-h5pw outright (both removed from the ignore file); the remaining four are documented per-advisory as blocked on the hackney 4.x major (httpoison requires ~> 1.17); and Projecthub.Github.PathSegment.validate/1 now rejects control characters, slashes and query/fragment markers in owner/name/login before the Tentacat calls.

adbatista (dismissed):

PR #1176, opened 2026-08-13
chore(periodic-scheduler): bump tackle to v0.4.0

loadez:


adbatista (dismissed):

skipi:
The tackle bump itself is a clean two-line tag flip. The new `.mix-audit.txt` entry is the problem: hackney is on a live runtime path in this service, and the stated provenance is wrong.

The file says hackney "has zero call sites anywhere in this service" and is "a fully unused transitive dependency pulled in via the grpc dependency chain". The first clause is true of first-party `lib/` code — there are genuinely no HTTPoison/Tesla/`:hackney` calls there. But "fully unused" does not follow, and grpc is not the path.

- `grpc 0.5.0` has no dependency edge to hackney at all. It pulls cowboy, cowlib and grpc_gun; grpc_gun 2.0.1 pulls only cowlib.
- The live path is timex → tzdata → hackney: `mix.exs:49` `{:timex, "~> 3.7"}` → timex 3.7.11 → tzdata 1.1.1 → hackney 1.18.1.
- tzdata 1.1.1 defaults to `autoupdate: :enabled` with `http_client: Tzdata.HTTPClient.Hackney` (`deps/tzdata/mix.exs:43-49`). `Tzdata.App` starts `Tzdata.ReleaseUpdater` whenever autoupdate is not `:disabled` (`tzdata_app.ex:12-15`), which fires 3 seconds after boot and every 5 hours thereafter (`release_updater.ex:13,17`), issuing `GET https://data.iana.org/time-zones/tzdata-latest.tar.gz` with `follow_redirect: true` (`data_loader.ex:7,11`) through `:hackney.get/4` and `:hackney.head/4` (`http_client/hackney.ex:9-17`). `:tzdata` declares `mod: {Tzdata.App, []}`, so it boots with the release.
- Nothing in this service disables it. There is no `:tzdata` configuration in `config/config.exs`, `dev.exs`, `prod.exs`, `runtime.exs`, `test.exs`, or anywhere in `lib/`.
- There is also a second rooted path: `mix.exs:54` path-depends on `definition_validator`, whose own `mix.exs:28` declares `{:httpoison, "~> 2.0"}` → httpoison 2.1.0 → hackney. So the httpoison entry in the lockfile is not stale.

Concretely, three of the six suppressed advisories sit on that live path: `GHSA-9fm9-hp7p-53mf` (connection-pool leak), `GHSA-gp9c-pm5m-5cxr` (no timeout on the post-handshake TLS upgrade), and `GHSA-vq52-99r9-h5pw` (SSRF via redirect target) — the last landing squarely on that `follow_redirect: true`. The remaining three need attacker-controlled URL input, which a hardcoded IANA endpoint does not supply, so those are defensible as written once the reasoning names the right dependency.

Simplest fix that makes the whole entry honest: add `config :tzdata, :autoupdate, :disabled`. `projecthub/config/runtime.exs:6` already does exactly this. Then hackney really is unused at runtime and the suppression stands on its own.

One related observation, probably worth its own issue rather than blocking here. Eight services in this repo share the pattern:

```elixir
if System.get_env("TZDATA_DATA_DIRECTORY") != nil do
  config :tzdata, :data_dir, System.get_env("TZDATA_DATA_DIRECTORY")
else
  config :tzdata, :autoupdate, :disabled
end
```

Setting a writable data directory therefore silently re-enables the network autoupdater — the two settings are mutually exclusive for no apparent reason. `TZDATA_DATA_DIRECTORY` is not set in any manifest or Dockerfile in this repo, so the safe branch runs today, but any operator who sets it turns outbound hackney traffic back on across all eight services at once. This service is the outlier that has neither branch.

PR #1180, opened 2026-08-14
Feat(front, github-notifier, periodic-scheduler, projecthub-rest-api, projecthub, public-api, repository-hub) Skip GitHub commit-status posts for Taks triggers

adbatista:
## 📝 Description
Adds two checkboxes to the task form 
  - "Don't send status checks for scheduled runs" and 
  - "Don't send status checks for Run now"

This lets a task silence the commit statuses it would otherwise post to your Git provider. Both default to unchecked, so every existing task keeps reporting; a checked box covers every pipeline that task starts with that trigger, reruns included, while pipelines started by a push, a pull request, or the API are never affected. 

The flags live on the task in periodic_scheduler and are honored by repository_hub's build status guard, which resolves suppression under the row lock that already serializes delivery for a check, so a decision that changes mid-pipeline can't leave a status stuck pending.

## ✅ Checklist


loadez:
Looks solid overall, the trigger enumeration and both-direction tests check out. Two things worth a look before merge:

1. Flag-preservation is handled on the REST update path but not the projecthub gRPC `ProjectService.Update` path. Since proto3 bools can't tell "unset" from "false", a caller sending non-empty `tasks` without re-stating the notification flags would reset them. REST pre-resolves the flags so it's fine; is there any other caller that passes `tasks` directly? If REST is the only one, a short test or comment asserting that would cover it, otherwise the merge could move into `attempt_update`.

2. `BuildStatusGuard.decide/4` now checks `live_lease?` before `stale_pending?`, changing the earlier ordering. In the mid-flight case it returns `:busy` (retry) where it previously skipped cleanly. Not incorrect, but it's a behavior change the moduledoc doesn't mention, plus an extra busy metric/log. Worth a note either way.

PR #1187, opened 2026-08-20
feat(front): add Turbo Drive SPA navigation

MathiasWP:
Replaces full-page MPA navigation with Turbo Drive, behind the `:turbo_navigation` feature flag (off by default). Our organization **kvist** would love to be a tester for this functionality!

## Changes

- **CSP** — adds `<meta name="csp-nonce">`. Turbo reads the *first* such tag, which stays the originally loaded page's, so the per-request nonce keeps matching the enforced CSP header. No CSP weakening needed.
- **Bundle** — `app.js` moves from `<body>` to `<head defer>`. Turbo re-executes every body script per visit, which would re-download the 10.9 MB bundle each navigation.
- **Init lifecycle** — `App.run()` splits into `runOnce` / `runPage` / `teardownPage`. Turbo replaces the `<body>` element, so `body`-delegated handlers must re-bind while `window`/`document` ones would otherwise stack.
- **Leak fixes** — `Pollman.stop()` only set a flag while its recursive `setTimeout` kept rescheduling, and `Timer` stored no interval handle. Both leaked per visit and already affected the existing `App.run()` callers.
- **Opt-outs** — workflow editor and job logs use `turbo-visit-control: reload`; they own Monaco/ace and log-streaming state.

## Scope limits

- `Turbo.setFormMode("off")` — link navigation only. Turbo requires form responses to redirect, and many controllers re-render a 200 on validation failure.
- `turbo-cache-control: no-preview` — previews don't fire `turbo:load`, so they'd paint a page whose per-page init hadn't run. Back/forward still uses the cache.

## Testing

`lint.js`, `compile.ts`, `test.js` (468 passing) and the non-browser Elixir suite (1813 tests, 0 failures) are green. The Wallaby browser suite has not been run yet.

### DIFF
--- a/front/assets/js/app.js
+++ b/front/assets/js/app.js
@@ -559,6 +559,7 @@ export var App = {
   teardownPage: function () {
     Pollman.stop();
     Timer.stop();
+    DiagramDrag.stop();
 
     if (window.FaviconUpdater) {
       window.FaviconUpdater.stop();
--- a/front/assets/js/app.js
+++ b/front/assets/js/app.js
@@ -80,6 +80,7 @@ import { DeploymentTargets } from "./deployments";
 
 import { Features } from "./features";
 import { Overlay } from "./overlay";
+import { TurboNavigation } from "./turbo_navigation";
 import { RoleForm } from "./roles/role_form.js";
 
 var ace = require('brace');
@@ -492,11 +493,40 @@ export var App = {
       config: InjectedDataByBackend.ReportConfig,
     })
   },
-  // App.run() is invoked at the bottom of the body element
-  run: function () {
+  //
+  // Bootstrap that must happen exactly once per document.
+  //
+  // Everything here binds to window, document or the custom element registry.
+  // Turbo keeps all three alive across visits, so re-running this would stack
+  // a duplicate scroll handler, click handler or posthog client per navigation.
+  //
+  runOnce: function () {
+    Overlay.init();
+
+    defineTimeAgoElement()
+    managePageHeaderShaddows()
+    enableMagicBreadcrumbs()
+    maybeEnablePosthog()
+
+    window.Tippy = Tippy;
+
+    $(document).on("click", ".x-select-on-click", function (event) {
+      event.currentTarget.setSelectionRange(0, event.currentTarget.value.length);
+    });
+
+    window.addEventListener("load", initPylonChatDraggable, { once: true });
+  },
+
+  //
+  // Bootstrap that has to happen for every page, including each Turbo visit.
+  //
+  // These either read page specific data out of InjectedDataByBackend, bind to
+  // elements in the current body, or - like Notice - delegate off document.body,
+  // which Turbo replaces wholesale on every render.
+  //
+  runPage: function () {
     Features.init(InjectedDataByBackend.Features || {});
 
-    Overlay.init();
     if (InjectedDataByBackend.InitialPlan) {
       TrialOverlay({
         dom: document.getElementById("trial-overlay"),
@@ -504,17 +534,10 @@ export var App = {
       })
     }
 
-    defineTimeAgoElement()
-    managePageHeaderShaddows()
-    enableMagicBreadcrumbs()
-    maybeEnablePosthog()
-
-
     if (InjectedDataByBackend.JumpTo !== undefined) {
       window.jumpTo = JumpTo.init();
     }
 
-    window.Tippy = Tippy;
     Tippy.defaultTip('[data-tippy-content]');
     Tippy.otherDefaultTip('.default-tip');
     Tippy.defaultDropdown('.js-dropdown-menu-trigger');
@@ -523,16 +546,24 @@ export var App = {
 
     window.Notice.init();
 
-
-    $(document).on("click", ".x-select-on-click", function (event) {
-      event.currentTarget.setSelectionRange(0, event.currentTarget.value.length);
-    });
-
     for (const el of document.querySelectorAll('[data-hotkey]')) {
       install(el);
     }
+  },
 
-    window.addEventListener("load", initPylonChatDraggable, { once: true });
+  //
+  // Releases the page that is about to be swapped out by Turbo. Modules that
+  // own timers, polling loops or editor instances have to be stopped here or
+  // they keep running against a body that no longer exists.
+  //
+  teardownPage: function () {
+    Pollman.stop();
+    Timer.stop();
+
+    if (window.FaviconUpdater) {
+      window.FaviconUpdater.stop();
+      window.FaviconUpdater = null;
+    }
   }
 };
 
@@ -622,7 +653,29 @@ function maybeEnablePosthog() {
   }
 }
 
-App.run()
-if (InjectedDataByBackend.JS != "" && InjectedDataByBackend.JS !== undefined) {
-  App[InjectedDataByBackend.JS]();
+//
+// Runs the per-page init function named by the backend, if there is one.
+// InjectedDataByBackend.JS is repopulated by _scripts.html.eex on every render,
+// so this picks up the new page's entry point after each Turbo visit.
+//
+function dispatchPageInit() {
+  const name = InjectedDataByBackend.JS;
+
+  if (name === undefined || name === "" || typeof App[name] !== "function") {
+    return;
+  }
+
+  App[name]();
 }
+
+App.runOnce()
+
+TurboNavigation.start({
+  onPageLoad: function () {
+    App.runPage();
+    dispatchPageInit();
+  },
+  onPageTeardown: function () {
+    App.teardownPage();
+  },
+})
--- a/front/assets/js/app.js
+++ b/front/assets/js/app.js
@@ -81,6 +81,7 @@ import { DeploymentTargets } from "./deployments";
 import { Features } from "./features";
 import { Overlay } from "./overlay";
 import { TurboNavigation } from "./turbo_navigation";
+import { islandRoot, unmountIslands } from "./preact_islands";
 import { RoleForm } from "./roles/role_form.js";
 
 var ace = require('brace');
@@ -96,7 +97,7 @@ window.SelfHostedAgents = SelfHostedAgents
 export var App = {
   agents: function () {
     Agents({
-      dom: document.getElementById("agents-app"),
+      dom: islandRoot(document.getElementById("agents-app")),
       config: InjectedDataByBackend.AgentsConfig,
     })
   },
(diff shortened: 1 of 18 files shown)


MathiasWP:
@loadez as mentioned in the PR description, we'd love to test this in our organisation in Semaphore! If we got this then i can follow up any errors that may appear, helping with making this reliable before it could be released to everyone!

skipi:
Thanks for this — the flag-gating design is right, and the `Pollman`/`Timer` fixes are real bugs that already affected the existing `App.run()` callers, so those stand on their own merit.

I've pushed a commit to your branch (you have maintainer edits enabled) so the PR stays the place we work, plus a merge of `main` to get CI green. Two findings behind the commit:

1. `DiagramDrag.init()` runs inside `App.workflow_view()`, which now re-runs per visit. Its `mousemove`/`mouseup` handlers are bound to `window` and its cursor style tag to `document.head`, neither of which the body swap replaces, so each visit to a workflow page stacked another set with no removal — same class as the leaks you fixed, and `mousemove` is a hot path. Worth a look at the other per-page inits for the same shape; `SleepDetector` and `workflow_editor/layout.js` are both already covered by your `turbo-visit-control: reload` opt-outs.
2. The three Turbo `<meta>` tags rendered unconditionally. They're inert without Turbo, but our bar for shipping this is that an organization without the flag renders an *identical* page rather than an equivalent one, so they're now behind a `LayoutView.turbo_enabled?/1` helper that the head and the scripts partial share.

One clarification on the lazy import: the deferral does work for *evaluation* — I rebuilt with our esbuild options to confirm no `Session`, no `ProgressBar` and no custom elements when the flag is off — but not for download. `assets/build.js` bundles without `splitting`, so Turbo lands inside `app.js` for everyone regardless. That's fine at roughly 0.4% of the bundle; I just adjusted the comment to match what the build actually does.

Could you give that commit a review?

On testing in kvist — that's exactly what we want, and it's the reason for the flag. One constraint on our side: `turbo_navigation` doesn't exist as a feature record yet, so it currently reads `false` everywhere with no way to flip it. We're sorting that out and doing a round on our internal environment first, and once that's done we can enable it for your organization. Worth flagging that `Turbo.setFormMode("off")` means link navigation only, so form flows won't be part of what you'd be exercising.

MathiasWP:
Reviewed the commit — both findings are correct and I'd have missed the `DiagramDrag` one. The `window` listeners plus the `document.head` style tag are exactly the shape that survives a body swap, and routing the teardown through `stop()` matches how `Pollman`/`Timer` work now. Gating the meta tags is the right call too; "identical, not equivalent" is a better bar than the one I shipped against. Thanks also for the correction on the lazy import — you're right that it only defers evaluation, and the amended comment says what the build actually does.

I did the audit you suggested and found one more, systemic rather than per-page: **Preact islands are never unmounted.**

The island components do clean up after themselves correctly — e.g. `insights/components/plot/plot.tsx` and `toolbox/popover.tsx` both return `removeEventListener` from `useEffect`. But those cleanups only run on unmount, and Turbo discards the body without telling Preact, so they never ran. Every visit to a page with an island left another live `resize`/`click` listener behind, holding a reference to detached DOM. Same shape as `DiagramDrag`, one level up.

Pushed `464c257b`:

- `preact_islands.js` — `islandRoot()` records a container at the point of mounting, `unmountIslands()` unmounts them all via `render(null, container)`, called from `teardownPage()` before the swap.
- Wrapped the 20 mount sites in `app.js`. Kept it at the `dom:` argument so the call sites still read as mounts.
- `preact_islands.spec.js` pins the behaviour the teardown depends on, including that the effect cleanup actually fires on unmount. Worth noting it needs `act()` to flush the effect first — without it the test passes for the wrong reason, since Preact defers `useEffect` and no cleanup is registered yet.

Rest of the audit came back clean: `TimeAgo` clears its interval in `disconnectedCallback`, `SleepDetector` is only reachable from job logs, and `workflow_editor/layout.js` and `monaco_code_editor.js` are behind the reload opt-out, as you said. `account/profile_management.js` binds `DOMContentLoaded` at module top level, which would never re-fire under Turbo — but it isn't imported anywhere, so it's dead code rather than a bug here.

One thing on `f26319ef`: seeding in `setup` also covers "is false for an organization the feature was never provisioned for", so that test ran with the feature existing-but-hidden and became a duplicate of the one above it. Since an absent feature record is precisely the state every organization is in today, I split the block — the two override tests keep your `setup`, and the unprovisioned ones now genuinely have no record. Same fix as yours otherwise; I'd hit the identical failure locally before I saw your push.

Green locally: `lint.js` 0 errors, `compile.ts` 0 errors, `test.js` 475 passing, Elixir 1817 tests / 0 failures. Worth flagging that `make test.ex` prefixes `mix test` with `-`, so a failing run still exits 0 — the two failures in your commit were only visible in the JUnit report, which is how I found them.

Understood on `turbo_navigation` not existing as a feature record yet, and no rush — happy to wait for your internal round. Link-only navigation is fine for what we'd be exercising; kvist spends most of its time moving between project, workflow and job pages, which is exactly the path this changes. The Wallaby suite still hasn't been run against any of this, so that's the gap I'd most want closed before it goes near a real organization.

MathiasWP:
CI is red on `464c257b` with "1 test failed". I chased it down — it looks like pre-existing flakiness in the browser suite rather than anything on this branch.

I couldn't read the failing test name from CI (the pipeline is in your org, so `pipeline_jobs` is out of scope for my token and the GitHub status only carries the one-line summary), so I reproduced locally.

**On this branch:** `test/browser` → 186 tests, 1 failure — `Front.Browser.WorkflowEditor.BlocksTest` "editing prologue commands", `Wallaby.StaleReferenceError` at `blocks_test.exs:62` on `Editor.select_first_block()`.

That one worried me, because the workflow editor is exactly where moving `app.js` to `<head defer>` changes when JS boots relative to render, and a stale reference means the DOM got replaced underneath the query. That timing shift applies with the flag off too, so it would have been the "identical, not equivalent" problem again. But it doesn't hold up:

| Scope | Result |
|---|---|
| `blocks_test.exs` alone (15 tests) | 0 failures |
| whole `workflow_editor/` directory (45 tests) | 0 failures |
| full `test/browser` (186 tests) | 1 failure |

**On `origin/main`, unmodified**, with assets rebuilt from main so the bundle wasn't contaminated: `test/browser` → 186 tests, **1 failure** as well, but a different test — `Front.Browser.ProjectSettings.ArtifactsTest` "deleting all policies on a project", `FunctionClauseError` on a nil session in `execute_query/2`.

Two full runs, two different single failures, one of them on untouched main. So the suite drops roughly one test per full run regardless of branch, and both failures are the stale-DOM/timing class that `AGENTS.md` already calls out. I'd expect a re-run of this PR's pipeline to come back green or fail somewhere unrelated — could you kick one off? I can't trigger it from my side.

Worth saying plainly: I reported this branch green earlier while having run the Elixir suite with `--exclude browser:true`, which is precisely the part CI runs. That was the gap I kept flagging as unclosed, and I should have closed it before pushing rather than after. The browser suite has now been run in full against this branch, and apart from the flake above it is clean at 185/186.

Separately, if it's useful, `ProjectSettings.ArtifactsTest` failing with a nil session on main suggests a session that has already been torn down being reused, which is a different bug from the stale-element flakes. Happy to open a separate issue for the two of them rather than let it ride on this PR.

MathiasWP:
Filed the browser suite flakiness as #1195, with both reproductions and the two root causes.

One correction to what I said above: the `ArtifactsTest` failure is not a torn-down session being reused. `remove_policy/3` does `Enum.at(all(section, ...), index)`, and `all/2` doesn't block the way `find/3` does — when the previously saved policy hasn't rendered yet the list is short, `Enum.at/2` returns `nil`, and that `nil` is what gets piped into `click/2`. Hence the `FunctionClauseError` with a nil first argument instead of a readable "element not found". Different cause from the stale-element race in the editor test, so the issue covers both separately.

This PR still needs a pipeline re-run when you get a chance.

MathiasWP:
CI came back green on `66585e56` — 3138 tests passed. Same code as the red run, so that confirms the earlier "1 test failed" was the browser suite flake rather than anything here.

`ui_turbo_navigation` is a better name, and matching the existing `ui_*` flags makes the surface obvious. Thanks for keeping the split describe block through the rename.

I've opened #1197 against main to fix the flakiness itself, so full-suite runs stop being ambiguous. It's test-only and independent of this PR. Worth noting the root cause turned out not to be what I put in #1195 — `all/2` was masking a lost write rather than causing it, details in the PR and a correction on the issue.

From my side this is ready whenever you are; the only thing still outstanding is provisioning `ui_turbo_navigation` so kvist can start exercising it.

loadez (approved):

skipi:
Quick heads-up: `main` here requires verified commit signatures, and every commit in a PR is checked — not just the merge result. Your two commits were unsigned, so the PR couldn't merge even with an approval.

I've re-signed them so this isn't holding you up — you're still the author on both, only the committer changed. **I force-pushed your branch to do it, so pull before you continue.**

Worth turning commit signing on locally, though: without it, every push you make needs someone here to rewrite the branch again before it can go in.

Nothing else is outstanding — the branch is current with `main` and CI is re-running on the re-signed commits. The islands fix was a good catch; it closes the same leak class as the `DiagramDrag` one.

skipi:
@MathiasWP this is merged, and `ui_turbo_navigation` is now enabled for kvist. The production deploy is rolling out as I write this — once it lands, allow ~10 minutes for the feature-flag cache to roll over and Turbo will be live for your organization only.

What you'll be exercising, so you know where to look and what isn't in scope yet:

- Link navigation only. Forms are deliberately off (`Turbo.config.forms.mode = "off"`), since plenty of our controllers re-render a 200 on validation failure.
- The workflow editor and job logs opt out via `turbo-visit-control: reload` — they own Monaco/ace and log streaming, so navigating to them is a normal full page load by design.
- Hover prefetching is off. Turbo 8 enables it by default; we opted out so adopting the library didn't silently add a speculative GET for every link a cursor crosses. Worth revisiting deliberately later.

Since Turbo keeps `window` and `document` alive across visits, the failure mode to watch for is things that accumulate rather than things that break outright — a page getting progressively slower the more you navigate, duplicated handlers firing twice, or a poller that keeps running after you've left the page. Repeated visits back and forth between the same two pages is the quickest way to surface that class of bug.

Thanks for the islands fix, and for sticking with this. Do shout if anything looks off.

MathiasWP:
Thank you @skipi! I've let the other developers know and they're really gassed! We'll post updates if we find anything that doesn't work as expected 🔥

MathiasWP:
Just tried it now @skipi, and it feels amazing! I'm so grateful you added this, the app feels a lot smoother now 😍

PR #1192, opened 2026-08-25
chore(github_hooks): bump mail to 2.9.1

dexyk:
## 📝 Description

Bumps the transitive `mail` gem to clear GHSA-mvxr-6m87-mv2q, an email address spoofing issue via malformed RFC 2047 encoded-words, flagged by bundler-audit.

| Gem | From | To | Fixes |
|---|---|---|---|
| `mail` (transitive, via actionmailer) | 2.9.0 | 2.9.1 | GHSA-mvxr-6m87-mv2q |

`mail` is loaded by the Rails actionmailer railtie, but github_hooks defines no mailers and neither sends nor parses email, so the vulnerable RFC 2047 address parser is not reachable; bumped to keep the transitive dependency patched.

Conservative `bundle lock --update mail --conservative`, so the diff is a single line (`mail 2.9.0` to `2.9.1`) and every other gem stays pinned. No Gemfile change (mail is transitive). 

## ✅ Checklist

### DIFF
--- a/github_hooks/Gemfile.lock
+++ b/github_hooks/Gemfile.lock
@@ -212,7 +212,7 @@ GEM
     loofah (2.25.2)
       crass (~> 1.0.2)
       nokogiri (>= 1.12.0)
-    mail (2.9.0)
+    mail (2.9.1)
       logger
       mini_mime (>= 0.1.1)
       net-imap


adbatista (approved):
I see that the app requires action_mailer: https://github.com/semaphoreio/semaphore/blob/main/github_hooks/config/application.rb#L7

but is it used?

the code looks nice, this comment is more to undertand, and if is not used we could remove from the require and not load `action_mailer` and its dependencies at all

dexyk:
@adbatista Good catch. Confirmed `action_mailer` is unused: no mailer classes and no `ActionMailer`/`Mail::`/`deliver_*` calls in `app` or `lib`.

The catch is that removing `require "action_mailer/railtie"` won't drop the `mail` gem or clear this finding. `mail` is pulled in by the `rails` meta-gem through both `actionmailer` and `actionmailbox` (both declare `mail (>= 2.8.0)`), so it stays in `Gemfile.lock` regardless of what we require at boot. So the 2.9.1 bump is still the fix here.

PR #1211, opened 2026-09-02
docs(self-hosted): document machine requirements for initialization jobs

DamjanBecirovic:
## Why

A customer building their own self-hosted agent image asked what software their machine needs to run initialization jobs. The docs don't answer this anywhere: the self-hosted install page lists no machine requirements, and the [pipeline initialization](https://docs.semaphoreci.com/using-semaphore/pipelines#init-job) reference says you can point init jobs at a self-hosted agent type without mentioning what that machine has to provide.

The non-obvious part is Erlang. `spc` shells out to a condition evaluator that runs on the Erlang VM, and it does so unconditionally — even for pipelines with no `change_in` or `when` expressions. An image with Git and Docker but no Erlang looks complete and fails every initialization job.

## What this adds

A new **Initialization job requirements** section (`self-hosted#init-requirements`) covering:

- Required software: Git (2.25+ recommended), Erlang/OTP, Bash, OpenSSH client, CA certificates, `which`, standard shell utilities
- Supported Erlang/OTP majors: **24, 25, 26, 27** — the toolbox ships one condition evaluator per major and selects by exact match, so anything outside that range leaves the machine without one
- What is *not* needed: Docker, Elixir, Git LFS
- An Ubuntu `apt-get` snippet and two verification commands

Plus two constraints that are easy to hit:

- **Erlang must be installed before the agent.** The toolbox resolves the OTP version once, at install time. Installing Erlang afterwards requires re-running `~/.toolbox/install-toolbox`
- **Initialization jobs are Linux-only** (x86_64 and arm64). The condition evaluator ships only in the Linux self-hosted toolbox, so self-hosted macOS and Windows agents can't run them

## Where the pointers go

The detail lives in one section; five entry points link to it so people find it from wherever they start.

| Page | Change |
|---|---|
| `using-semaphore/self-hosted.md` | New requirements section + a bullet in the limitations list |
| `using-semaphore/pipelines.md` | Link from the init-job agent reference |
| `using-semaphore/organizations.md` | Note where the init agent type is selected |
| `using-semaphore/self-hosted-install.md` | Note that this is needed *before* the agent |
| `reference/toolbox.md` | Explains why `spc` implies an Erlang dependency |

Applied to Cloud, CE and EE. `CE-1.4` / `EE-1.4` untouched, per `docs-contributing/VERSIONING.md`.

## Verification

- `npm run lint` — 437 files, 0 errors
- Confirmed every anchor linked to exists in all three editions, and each edition has an identical 7 references to `#init-requirements` (1 definition + 6 links)
- `npm run build` not run locally: local `node_modules` has `@docusaurus/core` 3.7.0 while the lockfile pins 3.9.1, so the committed `markdown.hooks` config is rejected. Pre-existing and unrelated to this diff — leaving it to CI


adbatista (dismissed):

DamjanBecirovic:
Added a follow-up commit after verifying the requirements against a real Semaphore agent with a testbox.

**Installing Erlang isn't sufficient — it has to be on the `PATH` of the shell that runs the job.**

On a Semaphore agent, Erlang is kerl-managed under `~/.kerl/installs/27.0` and `escript` only resolves in a login shell. Jobs run in a login shell, so this works in practice. But it means a self-hosted image that installs Erlang through a version manager (kerl, asdf, mise) which only activates for *interactive* shells will fail every initialization job while looking correctly provisioned.

It also produces a misleading diagnostic. A non-login check reports a false negative:

```
$ ssh <machine> 'which escript'
MISSING
$ ssh <machine> 'bash -lc "which escript"'
/home/semaphore/.kerl/installs/27.0/bin/escript
```

With `escript` off the `PATH`, `when` fails as `/usr/bin/env: 'escript': No such file or directory` — which reads like a missing-binary problem rather than a `PATH` one. The note includes the login-shell form of the check so nobody loses an afternoon to this.

Also confirmed while probing, supporting what the original commit says: the agent's toolbox directory still held `when_otp_24`, `when_otp_25` and `when_otp_26`, meaning the installer had moved `when_otp_27` into place based on the OTP version it detected. The exact-major-match selection is real, not inferred from reading the installer.

`npm run lint` clean (437 files, 0 errors). Rebased onto the `main` merge that landed on this branch in the meantime.

loadez (dismissed):

dexyk (approved):

PR #1212, opened 2026-09-02
fix(mcp_server): bump grpc to v1.83.2 for CVE-2026-84304

DamjanBecirovic:
## Why

The MCP Server `Check dependencies` job fails on `CVE-2026-84304` (**HIGH**) — gRPC-Go, fixed in 1.83.1. `mcp_server/go.mod` pinned `google.golang.org/grpc v1.82.1`.

This is currently blocking **any** pull request that touches the docs, not just MCP server changes. The block is gated on:

```
change_in('/mcp_server', ...) or change_in('/docs/docs', ...) or change_in('/docs/versioned_docs', ...)
```

because the service indexes and embeds the docs. It's the only failure on [#1211](https://github.com/semaphoreio/semaphore/pull/1211), a docs-only PR, and it does not clear on latest `main` — hence this standalone branch, so it can merge independently and unblock both.

## What changed

Only `mcp_server/go.mod` and `mcp_server/go.sum`.

Bumped to **v1.83.2**, the latest patch on the fixed minor line, rather than exactly v1.83.1 — same minor, so no new API surface, and it picks up subsequent patches. `go mod tidy` carried the transitive set along:

```
google.golang.org/grpc                   v1.82.1  -> v1.83.2
google.golang.org/genproto/.../rpc       (bumped)
golang.org/x/net                         v0.56.0  -> v0.58.0   (indirect)
golang.org/x/sync                        v0.21.0  -> v0.22.0   (indirect)
golang.org/x/sys                         v0.46.0  -> v0.47.0   (indirect)
golang.org/x/text                        v0.39.0  -> v0.41.0   (indirect)
```

## Verification

Everything ran in Docker on `go1.25.12` — nothing depends on a local toolchain.

Reproduced the failure and confirmed the fix with the same Trivy filesystem scan the security toolbox runs for Go (`--severity HIGH,CRITICAL --ignore-unfixed --scanners vuln`), and there is no `go-ignore-policy.rego` in `mcp_server`, so nothing is suppressed:

```
BEFORE (v1.82.1)  Total: 1 (HIGH: 1, CRITICAL: 0)
                  google.golang.org/grpc  CVE-2026-84304  HIGH  fixed  v1.82.1 -> 1.83.1
AFTER  (v1.83.2)  go.mod  gomod  0        -> trivy --exit-code 1 exits 0
```

`go build ./...` and `go vet ./...` both pass.

## Follow-up, not addressed here

Every other Go service in the repo is also below 1.83.1 and carries the same advisory. Their `Deployment Preconditions` blocks were skipped on the docs-only run because it doesn't touch their paths, so they aren't failing yet — but each will hit this on its next change:

```
repohub             v1.82.1
artifacthub         v1.82.1
bootstrapper        v1.79.3
encryptor           v1.79.3
loghub2             v1.79.3
public-api-gateway  v1.79.3
self_hosted_hub     v1.79.3
ee/velocity         v1.79.3
```

Deliberately out of scope to keep this independently mergeable. Worth a separate sweep.


adbatista (dismissed):

skipi (dismissed):

DamjanBecirovic:
Added `amqp091-go` v1.9.0 → **v1.13.0** for **CVE-2026-79921** (HIGH, DoS via oversized AMQP payloads, fixed in 1.13.0).

This advisory appeared after the grpc bump and fails the same `Check dependencies` job, so it was blocking this branch independently. The grpc fix is holding — the scan now reports only this one finding, and CVE-2026-84304 is gone.

**Why v1.13.0 and not v1.14.0:** 1.13.0 is the minimum that closes it. 1.14.0 is a further minor with no security benefit, and this already moves four minors on a library sitting in the message-queue path.

**The four-minor jump was verified, not assumed:**

- `go build ./...` and `go vet ./...` pass
- `go-tackle` — which wraps this library and is pinned to a Dec 2023 pseudo-version — compiles and runs against v1.13.0 unchanged. Under MVS it gets built against 1.13.0 whether or not it declares support, so this was the more plausible risk than the broker
- **Tested against a live RabbitMQ 3.13.7 broker**, matching production (3.13.7 / Erlang 26.2.5.5), driven through the same `tackle.NewPublisher` + `PublishWithContext` path `mcp_server/pkg/audit` uses. Connect, exchange/queue/bind, publish, consume and publisher confirms all succeed, with results identical to v1.9.0:

```
broker: product=RabbitMQ version=3.13.7
  client v1.9.0   -> roundtrip OK, publisher confirms ack=true
  client v1.13.0  -> roundtrip OK, publisher confirms ack=true
```

  This is the expected direction: amqp091-go speaks AMQP 0-9-1, which is unchanged between RabbitMQ 3.13 and 4.x. 4.x added native AMQP 1.0, but 0-9-1 is frozen, so a newer client against a 3.x broker is supported.

- Trivy, same flags as the security toolbox (`--severity HIGH,CRITICAL --ignore-unfixed`): **1 HIGH before, 0 after**, `--exit-code 1` exits 0

Diff is one line in `go.mod` plus 15 in `go.sum` — no transitive movement. All commands ran in Docker on go1.25.12.

**Still outstanding, deliberately not here:** five other Go services pin the same vulnerable `amqp091-go v1.9.0` — `artifacthub`, `loghub2`, `public-api-gateway`, `self_hosted_hub`, `ee/velocity`. They aren't failing yet because their blocks are skipped on changes that don't touch their paths.

dexyk (approved):

PR #1225, opened 2026-09-07
fix(zebra,self_hosted_hub): keep stopped jobs from reaching agents

adbatista:
## 📝 Description
`StopJob` leaves `assigned_job_id` in place and only the running-job sync path read `job_stop_requested_at`, so an agent whose job had been stopped before it asked for work was told to run it anyway — then fetched a payload whose secrets zebra had already scrubbed in the same update that finished the job, died exporting env vars, and reported the cancellation as a failure.

The hub now releases the assignment instead of dispatching, and zebra refuses to serve the payload of a finished job. `TaskFailFast` bulk-creates stop requests for every unfinished job in a task, which is why this arrived as bursts of several jobs failing across separate agents within a few hundred milliseconds.

## ✅ Checklist


adbatista:
@codex review

loadez (dismissed):

PR #1240, opened 2026-09-14
refactor(guard): gate /device auth-first flow on the device-auth marker cookie

loadez:


adbatista (dismissed):

guard/lib/guard/id/api.ex
```diff
@@ -1646,10 +1647,15 @@ defmodule Guard.Id.Api do
...
+    # Normalize the x-semaphore-user-id header to either a non-empty id or nil:
+    # a missing header and a blank value both mean "no user". This keeps
+    # assigns[:user_id] presence checks (e.g. the /signup logged_in flag)
+    # correct when the header is absent or blank.
     user_id =
-      conn
-      |> get_req_header("x-semaphore-user-id")
-      |> List.first()
+      case conn |> get_req_header("x-semaphore-user-id") |> List.first() do
+        id when is_binary(id) and id != "" -> id
```
adbatista: maybe, remove the List.first() and do 

`[id] when is_binary(id) and id != "" -> id`

loadez: Good call — applied. Matching on `[id]` drops the `List.first()` and asserts exactly one value: a blank, absent, or multi-valued header all normalize to `nil`.

adbatista (approved):

PR #1254, opened 2026-09-18
feat(artifacthub): delete a project's artifacts when the project is deleted

dexyk:
## Why

Deleting a project did not delete its artifacts. They survived the full 30 day grace period and only went when the project was hard-deleted, while our docs and the delete confirmation both say deleting a project removes everything.

## What

`artifacthub` now consumes the project lifecycle events `projecthub` already publishes, so no new call between the services was needed. Deleting a project marks its artifact storage and the existing bucket cleaner empties it.

The storage is kept for a grace period first, `ARTIFACT_PURGE_GRACE_PERIOD`, defaulting to three days. Restoring the project inside that window cancels the purge and the artifacts come back with it, which is what makes a mistaken delete recoverable. The storage, its bucket and its retention policy all survive, so a restored project can push again immediately. Only the contents go.

## Also fixed

- `DestroyArtifact` no longer overwrites a project's retention policies, which was why a restored project could come back set to delete its own artifacts after a day.
- The bucket cleaner's consumer had no retry, so it could lose its first connect to the broker and stay dead for the life of the pod, silently, leaving retention unapplied fleet-wide. Unrelated to this feature, but the feature cannot work without it.

## Rollout

Off unless `START_PROJECT_DELETION_WORKER=yes`, so it can ship and be enabled separately. One additive nullable column, no backfill. SaaS first, onprem later.

## Testing

Unit tests plus an end to end run on preprod: retention still applies unchanged, deleting keeps the artifacts through the window, restoring inside it brings them back intact, and nothing else was touched.

## ✅ Checklist
- [x] ~I have tested this change~


loadez (dismissed):

loadez:
Went through this end to end, including the projecthub side. With the flag off by default this is safe to merge, and the unconditional fixes (bucketcleaner consumer startup, retention rules no longer overwritten by `DestroyArtifact`, GCS delete errors surfaced) are welcome. Three things I'd like in this PR before it merges, because once the flag is on they decide whether a restored project keeps its artifacts:

**1. The comment in `projectdeletion/worker.go:9-12` is not accurate**
projecthub has no rule refusing a restore for an hour after deletion (`Project.restore`, and the gRPC `restore` handler, restore immediately). The design currently relies on that ordering, so the comment should go and the ordering needs handling in code.

**2. Restore should win regardless of event order**
The soft-delete and restore events land on separate queues with independent consumers, and `CancelArtifactPurge` is a no-op when no mark exists yet (`privateapi.go:133`). If the restore is handled first (consumers restarting, delete handler in retry), the delete marks the project afterwards and the purge runs on a live project. Both events carry a timestamp neither handler reads. Suggest persisting the restore time even when nothing is marked, and ignoring a soft-delete event that isn't newer.

**3. A test for restore-before-delete**
Covering the ordering above.

**Follow-ups (fine outside this PR):**
- `START_PROJECT_DELETION_WORKER` only gates the consumers. Turning it off during an incident leaves existing marks purging while restores go unseen. Worth documenting, or having the flag also stop purges.
- Purge mode is re-read per page (`cleaner.go:141`), so a chain that starts as retention and crosses into purge can clear the mark with early pages only retention-cleaned. `ClearPurgeMark` also runs unconditionally outside the transaction (`cleaner.go:148`). Fail-safe, data survives.
- Treat GCS `ErrObjectNotExist` on delete as success (`gcs_bucket.go:226-246`).
- `consumer_exited` fires on broker reconnects, so alerting on it would be noisy.
- `config.go:25-27` says scheduler and cleaner share a process. They're separate deployments, so the grace period has to be set on both.
- PR description says three days, the default is 24h (`config.go:16`).
- Optional, more robust version of #2: check with projecthub that the project is still soft-deleted before page 1 of a purge. That also covers a restore event lost at publish time or dead-lettered.

